PR #64, opened 2015-01-29
Adding logon as a service right

carpnick:
Add ability to change Logon As A service right at install time.


oleg-nenashev:
Thanks for the feature.
Is there a way to get such functionality using .NET or NuGet packages (we can use them since #62)? Usage of native libraries would be OK, but I'm quite aware about platform dependencies (minimal Win32 API version, etc.)

carpnick:
I can do some more research on this one to see if its possible.  Last time I looked, everywhere I looked this was only available to do from win32.

I am not the only one:
* https://bitbucket.org/splatteredbits/carbon/src/9d4e90e1ee3604e18e03eb6b0ba3f241463357a0/Source/Security/Privilege.cs?at=default
* http://www.codeproject.com/Articles/4863/LSA-Functions-Privileges-and-Impersonation

Several other examples found we can do this through GPO or C++.  WMI didnt seem to have native capability to perform this action.  From your comment, we can potentially move this functionality to a seperate Nuget package, but I dont think that is what you had in mind.  More of to limit Win32 correct?

If you want, I can go through the API I used, and do an assessment of minimum version of OS required, and can wrap the new feature in a conditional, only execute the code if minimum OS version is used.

oleg-nenashev:
Yes, I found many implementations of such code. I would vote to have a package for every code snippet on CodeProject :)

NuGet repo returns many projects, which implement the similar code (see examples below). Unfortunately, all NuGet projects have .NET 4.0+, but WinSW still supports .NET 2.0. Seems it's better to add a native implementation now and then replace it by an external library when the dependency gets updated (WinSW 2.0?).

It would be great if you check OS version dependencies and add checks if it is required.
It would be also useful to move native methods to <code>winsw.native</code> namespace.

Project examples:
* https://www.nuget.org/packages/Forge.WindowsServiceControl/
 * https://github.com/JZO001/Forge/blob/master/Forge.WindowsServiceControl/NativeMethods.cs
* https://www.nuget.org/packages/Simplify.WindowsServices/
 * Service engine: https://github.com/i4004/Simplify/tree/master/src/Simplify.WindowsServices

carpnick:
Ok will do.  Do you want to expose another parmeter that makes this change optional?  Make it an opt-in action?  If so any ideas on naming?

carpnick:
Also I can easily enough put this code in a Nuget package for .NET 2.0 and put it in a seperate repostiory

oleg-nenashev:
> Ok will do. Do you want to expose another parameter that makes this change optional? Make it an opt-in action? If so any ideas on naming?

I don't see a strong need in it. There could be a potential security issue, but Windows Service Manager also does not ask admins to confirm the change of permissions if you configure the it from GUI. Probably, somebody from reviewers has a strong opinion on it.

* For interactive mode you can ask user about the action. 
* For config-driven sources I would propose something like <code>allowServiceLogon</code> under <code>serviceaccount</code>.

> Also I can easily enough put this code in a Nuget package for .NET 2.0 and put it in a seperate repostiory

It will require additional efforts on the generalization, maintenance, etc.
If there's a bug, there will be additional efforts for the issue fixing due to permissions, etc. I would like to move winsw to a GitHub organization at some point, but it makes sense to wait with it.

Probably, you could just create a separate library project within WinSW. Then we will be able to convert it to NuGet package (together with WinSW) when there's a real need

carpnick:
Added configuration integration and interactive support
Also I went ahead and integrated all the win32 calls in with the wrapper classes that exist today.

carpnick:
Also added conditional logic based on OS version

README.markdown
```diff
@@ -278,8 +278,11 @@ It is possible to specify the useraccount (and password) that the service will r
        <domain>YOURDOMAIN</domain>
        <user>useraccount</user>
        <password>Pa55w0rd</password>
+       <allowserviceaccountlogonright>true</allowserviceaccountlogonright>
```
oleg-nenashev: The example uses a confusing casing. All other XML entries are documented to be lowercase, so I propose to follow this notation. It may become important if someone contributes an XSD schema for the config file

carpnick: This is all lower case is it not?  Doesn't really matter since I will be renaming this anyways - just need to understand your comment.

oleg-nenashev: I see "aLLowserviceaccountLogonright" in my browser (Mozilla 35.0.1).
BTW the copy-paste to text editors shows all letters in the lowercase, so seems it's an issue with fonts

carpnick:
All commits completed based on current feedback.

carpnick:
All set.

carpnick:
Just need to squash.

carpnick:
Closing in favor of #70

PR #70, opened 2015-01-30
Adding logon as a service right to user account specified in configuration

carpnick:
See PR #64 for context and review.


oleg-nenashev:
Thanks a lot for your efforts.
I suppose it's the last change in winsw-1.17.
Hopefully I'll kick-off the release on Monday

carpnick:
Thanks for the timely review and comments

Issue #69, opened 2015-01-30
Migrate WinSW to log4net or NLog


oleg-nenashev:
It's a follow-up to #34 and other issues. The internal logging should be improved, so an external logging engine would be useful.

log4net has been selected.

DONE:
-  Integrate log4net into WinSW, migrate wrapper.log engine

TODOs:
- Migrate EventLogs to log4net
- Delegate exceptions handling to log4net
- Allows setting up log levels from ServiceDescriptor
- (?) Migrate STDERR/STDOUT logs to log4net (#87)
- (?) Support XML configurations of log4net loggers

oleg-nenashev:
NLog is more powerful than log4net, but log4net could be preferable for Java developers (e.g. @kohsuke )

oleg-nenashev:
I've integrated the basic implementation in #73
Other features will be submitted as separate pull requests

oleg-nenashev:
Created one follow-up, closing this wide issue

PR #42, opened 2014-06-18
WinSW Extensions + SharedDirectoriesMapper

oleg-nenashev:
This is a dev. version of WinSW, which adds the extension mechanism to WinSW project.
The implementation has not been finished yet, but any comments will be useful.

Action items left:
* Implement Logger and pass it to extension methods
* Extend Unit tests
* Add license headers
* Revert MVS version before the merge

PR #74, opened 2015-01-31
[WiP] - WinSW Extensions Engine

oleg-nenashev:
The change  adds the extension mechanism to WinSW project. It also splits WinSW into the Core library and ServiceWrapper projects, which are being merged into a single executable.

See previous review comments in #42

Action items left:
* Rework Logger and pass it to extension methods (after #69)
* Extend Unit tests
* <code>ExtenbsionPoint</code> class and <code>Extension</code> attributes
* <code>EnvironmentContributor</code> and <code>ParameterResolver</code> (processing of XML data, e.g. password decryption)
* (?) Add support of remote plugins loading


oleg-nenashev:
In the current implementation ILMerge generates the unsigned binary. Probably, I'll have to switch it to <code>MSBuild.Community.Tasks.ILMerge</code> with manual <code>sn</code> calls. Hopefully there's another option

oleg-nenashev:
@mpdeimos , @carpnick , @kohsuke 
I would like to merge changes into winsw-2.0 branch in order to continue the rework. If you agree that the project structure is OK, IMO it makes sense to continue the development of <code>ExtensionPoint</code>s and hooks within the main branch.

src/Core/WinSWCore/Extensions/ExtensionPointAttribute.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>
+    /// This attribute is used to identify extension points within the code
+    /// </summary>
+    /// <remarks>
+    /// Each extension point implements its own entry type.
+    /// </remarks>
+    class ExtensionPointAttribute
+    {
+    }
```
oleg-nenashev: Just a erroneous stub. Please ignore

mpdeimos:
Looks good imho. A small comment (that can also be fixed on main): The namespace winsw.Plugins should be winsw.Extensions or the ExtensionManager be renamed to PluginManager (other classes of course as well). I'd just keep one single term for extension.

oleg-nenashev:
Agreed, I'll see how to normalize the structure . According to https://github.com/kohsuke/winsw#developer-info, the binary compatibility of winsw-2.0 is not guaranteed til the first release.

In my opinion, both terms are required. I would follow the common approach (it's being used in Jenkins project):
* <code>Extension</code> - implementation of an <code>ExtensionPoint</code> from the core
* <code>ExtensionManager</code> implements the loading and further search of extension by their type (to be implemented in the future)
* <code>Plugin</code> - an independent library/package, which may contain multiple extension points (e.g. <code>SharedDirectoriesMapper</code> plugin)

In the future I would use an external solution for plugins management.

mpdeimos:
Terminology now makes sense.

For taking a look at best practices for extension loading (or also also complete drop in), maybe take a look at [Mono.Addins](https://github.com/mono/mono-addins)

oleg-nenashev:
@mpdeimos 
Thanks for the reference. I investigated it ~1,5 years ago, but selected a fast implementation. "In the future I would use an external solution for plugins management." actually was related to Mono.Addins (there's not so many options for .NET BTW)

<code>Mono.Addins</code> seems to address all use-cases. The license and platform requiremens work as well. I'll submit a prototype implementation to this PR

oleg-nenashev:
NuGet returns bad .NET dependencies for Mono.Addins. The oldest available NuGet version supports .NET 3.5.

https://github.com/mono/mono-addins/releases/tag/0.5 can be used with .NET 2.0, but I'm not use if it is a good idea. There're plenty of changes since this version.

carpnick:
Looks generally pretty good.  Should we consider using serialization and deserialization for XML parsing for extensions and core as part of this work?

I used something like this in the past to send additional information to extensions.  The core project would have a base type of MyAppConfig class.  The extensions would then inherit from the config class and add their own properties.  When it came time to deserialize the xml into the derived class, you could keep the xml config in memory to allow each extension to deserialize into their specific derivative to get additional complex configuration.  Done correctly, the original core project would not fail on the extra configuration during deserialization.

oleg-nenashev:
I was able to build <code>Mono.Addins</code> for .NET Framework 2.0 after several changes (see https://github.com/oleg-nenashev/mono-addins/tree/mono-addins-1.2-net-2.0 ). I'm not sure if it works properly, but a custom build could be a solution till WinSW migrates to newer .NET versions.

In any case, seems it's required to split extensions to Factory and Extension instance classes.

>  Should we consider using serialization and deserialization for XML parsing for extensions and core as part of this work? ...
Yes, it makes sense. Serialization of all data types will be also mandatory if we support an unloading of extensions (I don't see such need for WinSW). In such case any external extension library should be stored in a separate assembly realm.

oleg-nenashev:
Mono.Addins prototyping does not cause significant changes in the project structure / extensions code. Merging this PR as is to continue the development of logging features.

I'll submit Mono.Addins integration as a separate pull requests when it's ready

PR #75, opened 2015-02-01
Add ability to set startup type to XML file

### LINKED ISSUE #23: Add "Startup type" configuration option
I need to create some services that will be started manually. Right now I have to manually (pun) go to services and change the startup type from automatic to manual myself. It would be better if there was such an option in the config xml.
Thanks.

carpnick:
Fixes #23

### DIFF
--- a/ServiceDescriptor.cs
+++ b/ServiceDescriptor.cs
@@ -5,6 +5,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Reflection;
 using System.Xml;
+using WMI;
 
 namespace winsw
 {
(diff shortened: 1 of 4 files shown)


ServiceDescriptor.cs
```diff
@@ -394,6 +395,31 @@ public string Description
...
+            {
+                var p = SingleElement("startmode", true);
+                if (p == null) return StartMode.Automatic;  // default value
+                try
+                {
+                    return (StartMode)Enum.Parse(typeof(StartMode), p, true);
+                }
+                catch
+                {
+                    Console.WriteLine("Start mode in XML must be one of the following:");
```
oleg-nenashev: Console IO from utility class is no good.
In winsw-2.0 it makes sense to replace it by a SEVERE error and post it in the console via log4net

carpnick: Agreed.  Since log4net wasnt in scope, wasnt sure how you wanted that handled.

PR #76, opened 2015-02-01
Fix needing to call install command from same directory.

### LINKED ISSUE #36: Service won't start if 'install' command is run from a different directory
It seems that if you run `winsw.exe install` from outside the directory where it is located, you get an error when trying to start the service.

Example:

With this directory structure:
```
c:\app\
c:\app\service\winsw.exe
c:\app\service\winsw.xml
```

And this `winsw.xml` file:
```xml
<service>
  <id>app</id>
  <name>app</name>
  <description>app</description>
  <executable>cmd</executable>
  <arguments>/D /C ping -t localhost</arguments>
</service>
```

If I install the service from I outside the directory: 
```
C:\app>service\winsw.exe install
``` 
and I try to start it: 
```
C:\app>service\winsw.exe start
```

I get this error:
```
WMI.WmiException: UnknownFailure
   at WMI.WmiRoot.BaseHandler.CheckError(ManagementBaseObject result)
   at WMI.WmiRoot.InstanceHandler.Invoke(Object proxy, MethodInfo method, Object[] args)
   at WMI.Win32ServiceProxy.StartService()
   at winsw.WrapperService.Run(String[] _args)
   at winsw.WrapperService.Main(String[] args)
```

If install it from its own folder: 
```C:\app\service>winsw.exe install```

I can run `winsw.exe start` from any folder and the service will start just fine.

Here's my windows info:
```
OS Name:                   Microsoft Windows 7 Enterprise
OS Version:                6.1.7601 Service Pack 1 Build 7601
```

Thanks for making such a useful tool!

carpnick:
Fixes  #36

### DIFF
--- a/ServiceDescriptor.cs
+++ b/ServiceDescriptor.cs
@@ -36,8 +36,7 @@ namespace winsw
                 // this returns the executable name as given by the calling process, so
                 // it needs to be absolutized.
                 string p = Environment.GetCommandLineArgs()[0];
-                return Path.Combine(AppDomain.CurrentDomain.BaseDirectory, p);
-
+                return Path.GetFullPath(p);
             }
         }
 


carpnick:
Since GetCommandLineArgs()[0] gives the EXE name.  It also includes the folder if calling from outside the directory.  

Solution:
Make C# handle the heavy lifting to figure it out. 

Tested inside and outside directory with #36 example config and was able to get service started.

oleg-nenashev:
There can be an impact on other commands like <code>restart!</code>
Some manual testing is required after the merge

PR #77, opened 2015-02-01
Fix null exception and throw a FileNotFoundException if xml file is neve...

### LINKED ISSUE #44: No check of the configuration file existance in ServiceDescriptor initialization
Reproduction steps:
* Launch winsw in a folder without ${EXECUTABLE_NAME}.xml

Log:
> ..\tmp>winsw.exe install 
> System.ArgumentNullException: Value cannot be null.
> Parameter name: path1
>   at System.IO.Path.Combine(String path1, String path2)
>   at winsw.ServiceDescriptor..ctor()
>   at winsw.WrapperService.Run(String[] _args)
>   at winsw.WrapperService.Main(String[] args)

carpnick:
...r found

Fixes #44

### DIFF
--- a/ServiceDescriptor.cs
+++ b/ServiceDescriptor.cs
@@ -45,18 +45,24 @@ namespace winsw
         {
             // find co-located configuration xml. We search up to the ancestor directories to simplify debugging,
             // as well as trimming off ".vshost" suffix (which is used during debugging)
+            //Get the first parent to go into the recursive loop
             string p = ExecutablePath;
             string baseName = Path.GetFileNameWithoutExtension(p);
             if (baseName.EndsWith(".vshost")) baseName = baseName.Substring(0, baseName.Length - 7);
+            DirectoryInfo d = new DirectoryInfo(Path.GetDirectoryName(p));
             while (true)
             {
-                p = Path.GetDirectoryName(p);
-                if (File.Exists(Path.Combine(p, baseName + ".xml")))
+                if (File.Exists(Path.Combine(d.FullName, baseName + ".xml")))
                     break;
+
+                if (d.Parent == null)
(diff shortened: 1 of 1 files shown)


ServiceDescriptor.cs
```diff
@@ -45,18 +45,24 @@ public ServiceDescriptor()
...
+            DirectoryInfo d = new DirectoryInfo(Path.GetDirectoryName(p));
             while (true)
             {
-                p = Path.GetDirectoryName(p);
-                if (File.Exists(Path.Combine(p, baseName + ".xml")))
+                if (File.Exists(Path.Combine(d.FullName, baseName + ".xml")))
                     break;
+
+                if (d.Parent == null)
+                    throw new FileNotFoundException("Unable to locate "+baseName+".xml file within executable directory or any parents");
```
oleg-nenashev: The message is incorrect.
The code searches across all upper directories. No idea why it is required, but seems this behavior should be retained.

carpnick: This is using your same behavior but getting rid of your null issue.

The loop first checks for the file in directory.fullname with the basename, and if it doesnt exist, the directoryinfo object parent property returns null if you have searched up to C:\

See https://msdn.microsoft.com/en-us/library/system.io.directoryinfo.parent(v=vs.80).aspx

Basically your loop still redefines your string directory going from directory to directory.  But there is a potential you could search to c:\, still not find the xml file.  This is the root cause of your current null issue.  If you try to  find DirectoryName of C:\ you get null

oleg-nenashev: Sorry. My message was about the error message only, but I have not read the entire message 
"or any parents" was not visible in GitHub code viewer (and I have not noticed a scrollbar)

Issue #80, opened 2015-02-01
Maven Build flow should pick releases from GitHub or NuGet


oleg-nenashev:
A correct Assembly version is required to track dependencies and to show it in CLI.

TODOs:
- Manually fix the version for winsw-1.x
- Adjust build flows to check/update Assembly info and pow.xml

Required for #78

oleg-nenashev:
The bug has been fixed in winsw-1.17
For winsw-2.0 it's just an improvement case

mrjoel:
This was updated for 1.18, however the 1.19 release still misreports 1.18 as the assembly version...

oleg-nenashev:
Fixed in 2.19.1 BTW it's still a risk

oleg-nenashev:
So the new flow has been merged. Maven release attempts will fail if the version differs from the WinSW release

Issue #46, opened 2014-06-24
WinSW does nothing on unsupported option in args[0]


oleg-nenashev:
Proposed behavior: Print help and fail with a non-null error code

NextTurn:
@oleg-nenashev This has been fixed.

PR #81, opened 2015-02-01
Add "version" and "help", handle unsupported commands

### LINKED ISSUE #78: Add "help" and "version" commands
Currently, there's no option to get a help info from the command line.

Related to #46

oleg-nenashev:
This rework minimally impacts the project structure.
Command line parsing and management will be improved in winsw-2.0

Resolves #78 and #46
Related to #80

Signed-off-by: Oleg Nenashev <o.v.nenashev@gmail.com>

### DIFF
--- a/Main.cs
+++ b/Main.cs
@@ -10,6 +10,7 @@ using System.Threading;
 using Microsoft.Win32;
 using WMI;
 using ServiceType = WMI.ServiceType;
+using System.Reflection;
 
 namespace winsw
 {
@@ -28,15 +29,30 @@ namespace winsw
         private bool _orderlyShutdown;
         private bool _systemShuttingdown;
 
-        public WrapperService()
+        /// <summary>
+        /// Version of Windows service wrapper
+        /// </summary>
+        /// <remarks>
+        /// The version will be taken from <see cref="AssemblyInfo"/>
+        /// </remarks>
+        public static Version Version
         {
-            _descriptor = new ServiceDescriptor();
+            get { return Assembly.GetExecutingAssembly().GetName().Version; }
+        }
+
+        public WrapperService(ServiceDescriptor descriptor)
+        {
+            _descriptor = descriptor;
             ServiceName = _descriptor.Id;
             CanShutdown = true;
             CanStop = true;
             CanPauseAndContinue = false;
             AutoLog = true;
             _systemShuttingdown = false;
+        }
+
+        public WrapperService() : this (new ServiceDescriptor())
+        {          
         }
 
         /// <summary>
@@ -511,11 +527,11 @@ namespace winsw
         }
 
         // ReSharper disable once InconsistentNaming
-        public static void Run(string[] _args)
+        public static void Run(string[] _args, ServiceDescriptor descriptor = null)
         {
             if (_args.Length > 0)
             {
-                var d = new ServiceDescriptor();
+                var d = descriptor ?? new ServiceDescriptor();
                 Win32Services svc = new WmiRoot().GetCollection<Win32Services>();
                 Win32Service s = svc.Select(d.Id);
 
@@ -612,6 +628,7 @@ namespace winsw
                             }
                         }
                     }
+                    return;
                 }
                 if (args[0] == "uninstall")
                 {
@@ -627,16 +644,19 @@ namespace winsw
                             return; // it's already uninstalled, so consider it a success
                         throw e;
                     }
+                    return;
                 }
                 if (args[0] == "start")
                 {
                     if (s == null) ThrowNoSuchService();
                     s.StartService();
+                    return;
                 }
                 if (args[0] == "stop")
                 {
                     if (s == null) ThrowNoSuchService();
                     s.StopService();
+                    return;
                 }
                 if (args[0] == "restart")
                 {
@@ -653,6 +673,7 @@ namespace winsw
                     }
 
                     s.StartService();
+                    return;
                 }
                 if (args[0] == "restart!")
                 {
@@ -666,6 +687,7 @@ namespace winsw
                     {
                         throw new Exception("Failed to invoke restart: "+Marshal.GetLastWin32Error());
                     }
+                    return;
                 }
                 if (args[0] == "status")
                 {
@@ -675,6 +697,7 @@ namespace winsw
                         Console.WriteLine("Started");
                     else
                         Console.WriteLine("Stopped");
+                    return;
                 }
                 if (args[0] == "test")
                 {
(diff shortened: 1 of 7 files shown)


oleg-nenashev:
@carpnick 
Could you briefly review the PR?
Since I have not started winsw-1.17 yet, it makes sense to merge your bugfixes and this change.

Main.cs
```diff
@@ -612,6 +628,7 @@ public static void Run(string[] _args)
                             }
                         }
                     }
+                    return;
```
carpnick: For all of these return calls.  Is it worth doing something like the following.  Make it easier to follow what Main is doing?
image

Not sure if this is within scope of the PR

Main.cs
```diff
@@ -682,8 +705,23 @@ public static void Run(string[] _args)
                     wsvc.OnStart(args.ToArray());
                     Thread.Sleep(1000);
                     wsvc.OnStop();
+                    return;
+                } 
+                if (args[0] == "help")
```
carpnick: This should probably support \? and -? as well.  Common syntax across all console applications

Main.cs
```diff
@@ -711,5 +749,38 @@ private static string ReadPassword()
...
+            Console.WriteLine("- 'install'   - install the service to Windows Service Controller");
+            Console.WriteLine("- 'uninstall' - uninstall the service");
+            Console.WriteLine("- 'start'     - start the service (must be installed before)");
+            Console.WriteLine("- 'stop'      - stop the service");
+            Console.WriteLine("- 'restart'   - restart the service");
+            Console.WriteLine("- 'status'    - check the current status of the service");
+            Console.WriteLine("- 'test'      - check if the service can be started and then stopped");  
+            Console.WriteLine("- 'version'   - print the version info");
+            Console.WriteLine("- 'help'      - print the help info");
+        }
```
carpnick: What about restart! and redirect.  Redirect could go into optional additional parameters or something like that

Tests/winswTests/MainTest.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        [Test]
+        public void PrintHelp()
+        {
+            string expectedVersion = WrapperService.Version.ToString();
+            string cliOut = CLITestHelper.CLITest(new[] { "help" });
+            
+            StringAssert.Contains(expectedVersion, cliOut, "Expected that help contains " + expectedVersion);
+            StringAssert.Contains("start", cliOut, "Expected that help refers start command");
+            StringAssert.Contains("start", cliOut, "Expected that help refers help command");
```
carpnick: Might want to change this test to validate based on your error message.  This one is looking for start but error is for help.  Assume copy/paste issue.

Main.cs
oleg-nenashev: I want to rework all commands to <code>enum</code>s in any case. It makes sense to store help info and other such stuff near the classes. IMO it could be in winsw-2.0 scope.

Enums in C# are not so powerful compared to Java ones, but helper classes will resolve this issue.

Tests/winswTests/MainTest.cs
oleg-nenashev: Yes, there's a copy-paste bug. That's why I vote for enums :)

Main.cs
oleg-nenashev: agreed

Tests/winswTests/Util/CLITestHelper.cs
```diff
@@ -0,0 +1,116 @@
...
+using System.IO;
+using JetBrains.Annotations;
+using winsw;
+
+namespace winswTests.Util
+{
+    /// <summary>
+    /// Helper for WinSW CLI testing
+    /// </summary>
+    public static class CLITestHelper
```
carpnick: Nice idea.  to test a main method.  Should we be instead trying to get as much functionality out of Main as possible?  Allowing us to test  each call in a separate internal static method as an example.  We can then use the following attribute to expose to the test project https://msdn.microsoft.com/en-us/library/system.runtime.compilerservices.internalsvisibletoattribute%28v=vs.110%29.aspx

carpnick: Then test code can concentrate on just one piece of functionality at a time

Main.cs
carpnick: sorry wrong slash - /?  - Linux to windows conversion is hard.

Tests/winswTests/Util/CLITestHelper.cs
oleg-nenashev: I would avoid making major project structure changes in winsw-1.x
For such purpose I would propose to conduct such refactoring in https://github.com/kohsuke/winsw/tree/winsw-2.0 (it's mapped to https://github.com/kohsuke/winsw/pull/71)

carpnick: agreed.  Once the restructuring is complete - we can then also do the restructure of the test code.

carpnick:
Dont have time to continue testing.  But because you brought in a Nuget item.  This will be a breaking change to your userbase.  See Bin directory of  winsw:
image

The Executable will not work without the jetbrains dll now.  I suggest you remove that part from this PR or this entire thing goes over to 2.0.  Many people depend on this item with a one file solution only

carpnick:
Actually just retested - try removing that nuget package from WinSW - you should still be able to build

oleg-nenashev:
> The Executable will not work without the jetbrains dll now. 

Yes, my bad. I was going to integrate annotations to tests only, but then selected wrong checkboxes in NuGet manager. For me the binary works well even W/O DDLs in the folder, but it definitely makes sense to remove the reference. 

https://github.com/kohsuke/winsw/pull/74 integrates ILMerge support into the project

oleg-nenashev:
Seems everything has been fixed

packages.config
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="JetBrains.Annotations" version="8.0.5.0" targetFramework="net20" />
```
carpnick: I think you got rid of the Nuget package incorrectly.  If you opened references and deleted the dll reference, that does not remove it from Nuget.  You have to do that from the Manage Nuget dependencies dialog.  This is an artifact that says your WinSW project still requires, via Nuget JetBrains.Annotations

winsw.csproj
```diff
@@ -92,6 +92,7 @@
     </Content>
   </ItemGroup>
   <ItemGroup>
+    <None Include="packages.config" />
```
carpnick: Once done properly, this change should also disappear since this is  the only Nuget Package you are using.

oleg-nenashev: NuGet package won't be included to the output if there's no reference in the code.
I can remove the package dependency to keep the repo clian

oleg-nenashev:
@carpnick , thanks for the review. Are there any additional comments? If no, I'll aggregate commits and then merge the PR

carpnick:
Rest of it looks good.

PR #82, opened 2015-02-01
Better diagnosis in install and uninstall commands

### LINKED ISSUE #55: UnknownFailure exception when service already exists
When winsw executes an install, and the service already exists an UnknownFailure is thrown:

```
WMI.WmiException: UnknownFailure
  at WMI.WmiRoot.BaseHandler.CheckError(ManagementBaseObject result)
  at WMI.WmiRoot.ClassHandler.Invoke(Object proxy, MethodInfo method, Object[] result)
  at WMI.Win32ServicesProxy.Create(String, String, String, ServiceType, ErrorControl, StartMode, Boolean, String[])
  at winsw.WrapperService.Run(String[] _args)
  at winsw.WrapperService.Main(String[] args)
```

Steps to reproduce:
1. Install winsw service (`jenkins-slave.exe install`).
2. Stop the service (`sc stop jenkins`)
3. Reinstall the service (`jenkins-slave.exe install`).

I tried to rename the service, and it still throws this exception. This is most likely due to the modifying the display name instead of the service name which was a bit confusing.

Only workaround is to delete the service (`sc delete jenkins`) and then the service could be installed again.

oleg-nenashev:
Changes:
* install - Fail the installation if the service exists
* uninstall - Print warning if the service does not exist

Resolves #55

Signed-off-by: Oleg Nenashev <o.v.nenashev@gmail.com>

### DIFF
--- a/Main.cs
+++ b/Main.cs
@@ -545,6 +545,14 @@ namespace winsw
                 args[0] = args[0].ToLower();
                 if (args[0] == "install")
                 {
+                    // Check if the service exists
+                    if (s != null)
+                    {
+                        Console.WriteLine("Service with id '" + d.Id + "' already exists");
+                        Console.WriteLine("To install the service, delete the existing one or change service Id in the configuration file");
+                        throw new Exception("Installation failure: Service with id '" + d.Id + "' already exists");
+                    }
+
                     string username=null, password=null;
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
@christ66
Does it resolve the issue?

PR #92, opened 2015-04-04
[NO_MERGE] [winsw-2.0] - Integration of Mono.addins

oleg-nenashev:
The current version is not ready for the review. I'm migrating to another Windows laptop, hence it will take some time to deploy the environment and continue working on the feature


oleg-nenashev:
Well, I think the approach does not really make sense since it's too complicated to maintain the backported lib. Maybe I'll do it once the code gets migrated to .NET 4

oleg-nenashev:
It's not going to Fly in such way

PR #114, opened 2016-05-25
README: Escape allowservicelogon tags

r-murphy:
'allowservicelogon' in the note wasn't displaying due to the xml tags.


oleg-nenashev:
Thanks!

Issue #125, opened 2016-10-25
New plugin: Terminate runaway processes on startup [JENKINS-39231]


oleg-nenashev:
In Jenkins projects we have many users complaining that the slave/agent is "already connected", because they have a runaway slave/agent process. It happens when WinSW gets terminated without executing the process shutdown logic (force kill, JVM crash) or when WinSW fails to terminate the process.

As a part of WinSW 2.0, it would be great to create a logic, which...
- records PID of the created process to the disc
- performs status check of the previously spawned process upon restart
- terminates the runaway process if required

It can be done via WinSW 2 "plugin"

https://issues.jenkins-ci.org/browse/JENKINS-39231

oleg-nenashev:
Related Jenkins issues: https://issues.jenkins-ci.org/browse/JENKINS-39231

calebboyd:
This would be really slick if it solves the problem described in #105. We're running hundreds of installations and we see this happen frequently. Need any help here?

oleg-nenashev:
@calebboyd Well, any kinds of testing and reviews would be useful. I have created #133 with the change.

oleg-nenashev:
The pull request has been integrated towards 2.0

PR #129, opened 2016-11-24
Documentation refactoring towards 2.0

oleg-nenashev:
- [x] Decouple the documentation to several pages
- [x] Cleanup wording 
- [x] Add changelog
- [x] Ensure all internal and external links are correct
- [x] Integrate Wiki to close #132
- [x] Add some missing basic information about the installation flow

PR #133, opened 2016-11-26
[Issue #125] Runaway process termination on startup

oleg-nenashev:
Addresses #125. Changes summary:

- [x] `IWinSWExtension` now offers hooks for the wrapped process startup/termination
- [x] Some generic process management logic has been moved from `Main` to `ProcessHelper` in `WinSW Base`
- [x] New extension point - `Runaway Process Killer` with some basic tests

This change introduces an extension, which performs automatic termination of runaway processes on startup. Sample configuration:

```
﻿<?xml version="1.0" encoding="utf-8" ?>
<service>
  <id>SERVICE_NAME</id>
  <name>Jenkins Slave</name>
  <description>This service runs a slave for Jenkins continuous integration system.</description>
  <executable>%BASE%\sleep.bat</executable>
  <arguments></arguments>
  <logmode>rotate</logmode>

  <extensions>
	<!-- This is a sample configuration for the RunawayProcessKiller extension. -->
    <extension enabled="true" className="winsw.Plugins.RunawayProcessKiller.RunawayProcessKillerExtension" id="killOnStartup">
      <!-- Absolute path to the PID file, which stores ID of the previously launched process. -->
      <pidfile>%BASE%\pid.txt</pidfile>
      <!-- Defines the process termination timeout in milliseconds. This timeout will be applied multiple times for each child process. -->
      <stopTimeout>5000</stopTimeout>
      <!-- If true, the parent process will be terminated first if the runaway process gets terminated. -->
      <stopParentFirst>false</stopParentFirst>
    </extension>
  </extensions>
</service>

```

https://issues.jenkins-ci.org/browse/JENKINS-39231

### DIFF
--- a/src/Core/WinSWCore/Util/ProcessHelper.cs
+++ b/src/Core/WinSWCore/Util/ProcessHelper.cs
@@ -1,0 +1,109 @@
+﻿using log4net;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Management;
+using System.Text;
+
+namespace winsw.Util
+{
+    /// <summary>
+    /// Provides helper classes for Process Management
+    /// </summary>
+    /// <remarks>Since WinSW 2.0</remarks>
+    public class ProcessHelper
+    {
+        private static readonly ILog Logger = LogManager.GetLogger(typeof(ProcessHelper));
+
+        /// <summary>
+        /// Gets all children of the specified process.
+        /// </summary>
+        /// <param name="pid">Process PID</param>
+        /// <returns>List of child process PIDs</returns>
+        public static List<int> GetChildPids(int pid)
+        {
+            var searcher = new ManagementObjectSearcher("Select * From Win32_Process Where ParentProcessID=" + pid);
+            var childPids = new List<int>();
+            foreach (var mo in searcher.Get())
+            {
+                var childProcessId = mo["ProcessID"];
+                Logger.Info("Found child process: " + childProcessId + " Name: " + mo["Name"]);
+                childPids.Add(Convert.ToInt32(childProcessId));
+            }
+            return childPids;
+        }
+
+        /// <summary>
+        /// Stops the process.
+        /// If the process cannot be stopped within the stop timeout, it gets killed
+        /// </summary>
+        /// <param name="pid">PID of the process</param>
+        /// <param name="stopTimeout">Stop timeout</param>
+        public static void StopProcess(int pid, TimeSpan stopTimeout)
+        {
+            Logger.Info("Stopping process " + pid);
+            Process proc;
+            try
+            {
+                proc = Process.GetProcessById(pid);
+            }
+            catch (ArgumentException ex)
+            {
+                Logger.Info("Process " + pid + " is already stopped", ex);
+                return;
+            }
+
+            Logger.Info("Send SIGINT " + pid);
+            bool successful = SigIntHelper.SendSIGINTToProcess(proc, stopTimeout);
+            if (successful)
+            {
+                Logger.Info("SIGINT to" + pid + " successful");
+            }
+            else
+            {
+                try
+                {
+                    Logger.Warn("SIGINT to " + pid + " failed - Killing as fallback");
+                    proc.Kill();
+                }
+                catch (ArgumentException)
+                {
+                    // Process already exited.
+                }
+            }
+
+            //TODO: Propagate error if process kill fails? Currently we use the legacy behavior
+        }
+
+        /// <summary>
+        /// Terminate process and its children.
+        /// By default the child processes get terminated first.
+        /// </summary>
+        /// <param name="pid">Process PID</param>
+        /// <param name="stopTimeout">Stop timeout (for each process)</param>
+        /// <param name="stopParentProcessFirst">If enabled, the perent process will be terminated before its children on all levels</param>
+        public static void StopProcessAndChildren(int pid, TimeSpan stopTimeout, bool stopParentProcessFirst)
+        {
+            var childPids = GetChildPids(pid);
+
+            if (stopParentProcessFirst)
+            {
+                StopProcess(pid, stopTimeout);
+                foreach (var childPid in childPids)
+                {
+                    StopProcessAndChildren(childPid, stopTimeout, stopParentProcessFirst);
+                }
+            }
+            else
+            {
+                foreach (var childPid in childPids)
+                {
+                    StopProcessAndChildren(childPid, stopTimeout, stopParentProcessFirst);
+                }
+                StopProcess(pid, stopTimeout);
+            }
+        }
+
+        // TODO: Also move StartProcess methods once LogEvent()/WriteEvent() mess gets solved
+    }
+}
--- a/src/Core/WinSWCore/Util/XmlHelper.cs
+++ b/src/Core/WinSWCore/Util/XmlHelper.cs
@@ -13,7 +13,7 @@ namespace winsw.Util
         /// </summary>
         /// <param name="node">Parent node</param>
         /// <param name="tagName">Element name</param>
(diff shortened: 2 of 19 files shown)


oleg-nenashev:
Also created https://github.com/kohsuke/winsw/releases/tag/PR-133 with winsw.exe for evaluation

oleg-nenashev:
@reviewbybees

oleg-nenashev:
Plugin documentation will be delivered later once I get #129 integrated

src/Core/WinSWCore/Util/ProcessHelper.cs
```diff
@@ -0,0 +1,109 @@
...
+        private static readonly ILog Logger = LogManager.GetLogger(typeof(ProcessHelper));
+
+        /// <summary>
+        /// Gets all children of the specified process.
+        /// </summary>
+        /// <param name="pid">Process PID</param>
+        /// <returns>List of child process PIDs</returns>
+        public static List<int> GetChildPids(int pid)
+        {
+            var searcher = new ManagementObjectSearcher("Select * From Win32_Process Where ParentProcessID=" + pid);
```
jtnord: Does this still work with a limited access user?  (ie not running as SYSTEM but a locked down (non-admin) service account)

src/Core/WinSWCore/Util/XmlHelper.cs
```diff
@@ -13,7 +13,7 @@ public class XmlHelper
         /// </summary>
         /// <param name="node">Parent node</param>
         /// <param name="tagName">Element name</param>
-        /// <param name="optional">If otional, don't throw an exception if the elemen is missing</param>
+        /// <param name="optional">If optional, don't throw an exception if the elemen is missing</param>
```
jtnord: you missed the second typo on the line :)  (elemen**t**)

src/Plugins/RunawayProcessKiller/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("RunawayProcessKiller")]
+[assembly: AssemblyDescription("Kills runaway process on startup")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("")]
+[assembly: AssemblyProduct("RunawayProcessKiller")]
+[assembly: AssemblyCopyright("Copyright ©  2015")]
```
jtnord: 2015?

src/Plugins/RunawayProcessKiller/RunawayProcessKiller.csproj
```diff
@@ -0,0 +1,72 @@
...
+    <DebugType>pdbonly</DebugType>
+    <Optimize>true</Optimize>
+    <OutputPath>bin\Release\</OutputPath>
+    <DefineConstants>TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="log4net">
+      <HintPath>..\..\packages\log4net.2.0.5\lib\net20-full\log4net.dll</HintPath>
```
jtnord: NIT: does a user need to do more setup now to get log4net so they can compile this project?  (where can they obtain it from - is the README updated?)

src/Plugins/RunawayProcessKiller/RunawayProcessKillerExtension.cs
```diff
@@ -0,0 +1,120 @@
...
+                }
+                catch (FormatException e)
+                {
+                    Logger.Error("Invalid PID file number in '" + Pidfile + "'. The runaway process won't be checked", e);
+                    return;
+                }
+            }
+            else
+            {
+                Logger.Warn("The requested PID file '" + Pidfile + "' does not exist. The runaway process won't be checked");
```
jtnord: NIT: is this really a warning?  - seems like INFO as this is expected (at least for the first time you startup the service)

src/Plugins/RunawayProcessKiller/RunawayProcessKillerExtension.cs
```diff
@@ -0,0 +1,120 @@
...
+                proc = Process.GetProcessById(pid);
+            }
+            catch (ArgumentException ex)
+            {
+                Logger.Debug("No runaway process with PID=" + pid + ". The process has been already stopped.");
+                return;
+            }
+
+            // Kill the runaway process
+            ProcessHelper.StopProcessAndChildren(pid, this.StopTimeout, this.StopParentProcessFirst);
```
jtnord: This is very scary.... on windows PIDs can be reused...  and if this is due to a reboot after say patching the PID could depend on the order of startup of services - so you could (attempt) kill a very important system service.

There should be some form of check that the parent process was the one started (perhaps by setting a magic environment cookie - or at least looking like it is a java process aka Jenkins)

src/Test/winswTests/Extensions/RunawayProcessKillerTest.cs
```diff
@@ -0,0 +1,63 @@
...
+        [SetUp]
+        public void SetUp()
+        {
+            string testExtension = typeof (RunawayProcessKillerExtension).ToString();
+            string seedXml = "<?xml version=\"1.0\" encoding=\"utf-8\" ?>"
+                + "<service>                                                                                                        "
+                + "  <id>SERVICE_NAME</id>                                                                                          "
+                + "  <name>Jenkins Slave</name>                                                                                     "
+                + "  <description>This service runs a slave for Jenkins continuous integration system.</description>                "
+                + "  <executable>C:\\Program Files\\Java\\jre7\\bin\\java.exe</executable>                                               "
```
jtnord: this test is likely to fail on a CI system like Jenkins if this is needed - if its not - then should really be something like x:\ignored\not\used to avoid confusion.

src/Core/WinSWCore/Util/ProcessHelper.cs
oleg-nenashev: It used to work at least. The code has been just migrated to another place without change, so I do not expect differences

src/Plugins/RunawayProcessKiller/Properties/AssemblyInfo.cs
oleg-nenashev: Will fix it

src/Plugins/RunawayProcessKiller/RunawayProcessKiller.csproj
oleg-nenashev: No need to do additional steps. log4net is being merged into the executable using `ILMerge`. A user still gets a single binary

jtnord:
@oleg-nenashev generally the code looks good - but killing process trees because a PID was left in a file looks very very scary to me - and I fear you could end up killing very important services on a restart of the OS (due to the old PID being used by a different service after restart) at which point you could crash windows (if its a system critical process) and which point you get boot loops....

src/Plugins/RunawayProcessKiller/RunawayProcessKillerExtension.cs
oleg-nenashev: Agreed

oleg-nenashev:
Yes, I agree the process validation needs to be extended

src/Core/WinSWCore/Util/ProcessHelper.cs
jtnord: ok, missed that :o

oleg-nenashev:
@jtnord I have added the check of the Service Id environment of the process, which needs to be terminated

oleg-nenashev:
@reviewbybees done
Here we go

Issue #124, opened 2016-10-19
High CPU Usage on Stop with External Stop Command


jriffel:
I am using winsw with an external stop command.  I observed consistently that while the external stop command is executing the winsw process is taking nearly all available CPU.  This was causing me issues because the real winsw process struggles to shut down due to lack of CPU resources.  To be clear, this is the winsw process which is spawned in order to issue the stop, not the one currently running which is being stopped.

I tested versions 1.19, 1.18, and 1.12.  Version 1.12 does not have this issue and has all the features I need so I switched to 1.12.

Thanks.

pkeusem:
We have seen this problem as well.  Some debugging has lead me to a cause.  I am working with version 1.19.1.

There are actually two problems:

- First, the waithint and sleeptime properties of the service description are being ignored.  I have not been able  to figure out why yet but the end result is that the default value is always used.

- Second, the WaitHint and SleepTime properties are TimeSpans.  WaitHint is used by the SignalShutdownPending and SignalShutdownComplete methods, SleepTime is used by the WaitForProcessToExit method.  In all three cases, the Milliseconds property of the TimeSpan is used.  The problem with this is that Milliseconds only gets the Milliseconds component of the time span.  For example, if the time span is 10 seconds, Milliseconds is 0.  The default value for waithint is 15 seconds and the default value for sleeptime is 1 second.  This results in an effective waithint of 0 and and an effective sleeptime of 0.  So winsw is essentially stuck in a hard spin in a busy loop until the shutdown process exits.

Modifying the SignalShutdownPending and WaitForProcessToExit  methods to include minutes and seconds in the waithint and sleeptime calculations resolves the busy loop problem.

pkeusem:
OK, I have found fixes to both of the problems that I mentioned above:

- Using the TotalMilliseconds property of TimeSpan instead of Milliseconds resolves the problem of Milliseconds being zero.
- Replacing dom.FirstChild with dom in the SingleTimeSpanElement() invokes in WaitHint, SleepTime, ResetFailureAfter and StopTimeout resolves the problem of the waithint, sleeptime, resetfailure and stoptimeout config file property settings being ignored.

oleg-nenashev:
@pkeusem Thanks for diagnosing it. The first bullet definitely needs a fix, likely the second one as well

pkeusem:
OK, I have found fixes to both of the problems that I mentioned above:

[...]

PR #135, opened 2016-11-26
[Issue #124] - Prevent the CPU overutilization due to the usage of Milliseconds instead of TotalMsec

oleg-nenashev:
See #124 

@reviewbybees


oleg-nenashev:
@reviewbybees done

PR #140, opened 2016-11-30
Updates Documentation for Puppet Module

kenmaglio:


oleg-nenashev:
Thanks for moving the default guide! I think it's perfect for now. Once we are at risk of a significant unsync between docs, we can discuss better options

PR #155, opened 2016-12-08
How to use (Group) Managed Service Accounts

beatcracker:
Added example of configuring WinSW to use GMSA. Tested with Windows Server 2012 R2.


oleg-nenashev:
👍

PR #156, opened 2016-12-08
Handover Version control to AppVeyor jobs

oleg-nenashev:
In order to maintain Release and Snapshot versions on AppVeyor, I had to create two separate jobs on it. These jobs utilise the same config, but actually they only specify different version formats.

* Release job: `2.0.{build}`
* Snapshot & PR builder: 2.0-SNAPSHOT-build{build}

Maybe it's not the best practice

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,3 @@
-
-version: '2.0-SNAPSHOT-build{build}'
-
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Merging just to move forward with the release preparations.
If somebody has better ideas about handling of such use-case, I would appreciate it

Issue #146, opened 2016-12-04
Appveyor: Register webhooks


oleg-nenashev:
In order to get a real PR builder, an Appveyor web hook needs to be added to this repository. I have no permission to do it, hence I need some help from @kohsuke. I will ping him to get proper permissions.

oleg-nenashev:
Thanks @kohsuke, the pull request builder is working fine now

PR #165, opened 2016-12-14
Issue #146 - Improve logging of the Uninstall Operation.

oleg-nenashev:
This change does not change the behavior (return code, etc.) of the WinSW logic, but on the other hand it provides diagnostic information.

Since 2.0 is approaching, I would appreciate feedback regarding changing the default behavior. I would rather implement a `uninstall!` command, which stops the running service and fails horrible if the service cannot be deleted. But maybe it's more convenient to have a breaking change

@reviewbybees


oleg-nenashev:
@reviewbybees done
Merging as a maintainer

PR #166, opened 2016-12-20
Do not propagate exceptions from Process.Kill() if the process actually exits

iforapsy:
Sometimes we get an InvalidOperationException when we kill a process that is already dead and other times we get a Win32Exception. Let's ignore all exceptions from killing a process if it has exited.

Additionally, ArgumentException is the wrong exception to catch. I have never seen Process.Kill() throw that. It seems to be a bad copy and paste from line 50.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,6 @@ New versions with fixes may be released on-demand.
 ### Build Environment
 
 * IDE: [Visual Studio Community 2013](http://www.visualstudio.com/en-us/news/vs2013-community-vs.aspx) (free for open-source projects)
-* `winsw-key.snk` should be available in the project's root in order to build the executable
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
Agreed, it's reasonable

README.md
```diff
@@ -90,6 +90,6 @@ New versions with fixes may be released on-demand.
 ### Build Environment
 
 * IDE: [Visual Studio Community 2013](http://www.visualstudio.com/en-us/news/vs2013-community-vs.aspx) (free for open-source projects)
-* `winsw-key.snk` should be available in the project's root in order to build the executable
+* `winsw_key.snk` should be available in the project's root in order to build the executable
```
oleg-nenashev: thanks for catching it

Issue #85, opened 2015-02-01
Get rid of the deprecated property call


oleg-nenashev:
Main.cs:538

> Warning   1   'System.IO.FileStream.Handle' is obsolete: 'This property has been deprecated.  Please use FileStream's SafeFileHandle property instead.  http://go.microsoft.com/fwlink/?linkid=14202' C:\Users\nenashev\Documents\Jenkins\contrib\winsw\Main.c

oleg-nenashev:
The fix has been integrated towards 2.0

PR #167, opened 2016-12-21
Issue #85 - Use SafeFileHandle of the deprecated FileHandle in the /redirect mode

oleg-nenashev:
Solves #85


oleg-nenashev:
I confirmed that it works manually. A test case would be useful

PR #175, opened 2016-12-28
Fix #159 - Time span settings parsing

### LINKED ISSUE #159: TimeSpan properties of the service description are being ignored. I have not been able to figure out why yet but the end result is that the default value is always used.
It is a follow-up to the comment from @pkeusem  in #124

The proposed fix is:
> Prefixing the tagname with "//" in SingleTimeSpanElement() and replacing dom.FirstChild with dom in the SingleTimeSpanElement() invokes in WaitHint, SleepTime, ResetFailureAfter and StopTimeout resolves the problem of the waithint, sleeptime, resetfailure and stoptimeout config file property settings being ignored.

oleg-nenashev:
Adds tests to verify timespan field parsing correctness for the issue reported in #159 

- [x] Add a Utility class for building test configs
- [x] Add tests for all 4 TimeSpan fields
- [x] Fix #159 

@reviewbybees


oleg-nenashev:
CC @pkeusem and @jriffel
The tests pass correctly on my machine and the PR builder.
Could you please provide your configs in #159 ?

src/Core/WinSWCore/ServiceDescriptor.cs
```diff
@@ -126,16 +126,8 @@ private int SingleIntElement(XmlNode parent, string tagName, int defaultValue)
 
         private TimeSpan SingleTimeSpanElement(XmlNode parent, string tagName, TimeSpan defaultValue)
         {
-            var e = parent.SelectSingleNode(tagName);
```
oleg-nenashev: There is similar pattern for Int above, but it is not being used to parse config entries

src/Core/WinSWCore/ServiceDescriptor.cs
```diff
@@ -505,7 +497,7 @@ public TimeSpan WaitHint
         {
             get
             {
-                return SingleTimeSpanElement(dom.FirstChild, "waithint", Defaults.WaitHint);
```
oleg-nenashev: FirstChiled can be a comment or XML version definition...

oleg-nenashev:
Merging since I want to deliver another fix on the top of the test framework
@reviewbybees done (I've tried to get reviews)

Issue #126, opened 2016-10-25
Add support of authentication in download settings


oleg-nenashev:
This ticket was created for the download plugin, but currently the enhancements are being done on the level of the existing `download` command

nightman68:
Hi Oleg,

I’m currently investigating a Jenkins setup for our company. I have one Windows slave configured and when I upgraded recently to the new Jenkins release the Windows slave turned offline because the slave was to old. I found this project here on GitHub and it perfectly meets our requirements to update the slave.jar automatically before the service on the Windows slave is started.

In the configuration file for the service I used the option
`
<download from=“https://jenkins.corp/jnlpJars/slave.jar” to=“%BASE%\slave.jar”/>
`
but I received an authentication error. 

The Jenkins server is using the reverse proxy plugin for authentication in combination with Kerberos authentication via mod_aut_kerb. The connection to the master is working but the download fails.

My question is: how the download is working, I assume that it is Java based?! If yes, is there a possibility to pass the required Java options for GSSAPI authentication to the download?

Thanks!

oleg-nenashev:
@nightman68 
Hi Torsten,

> how the download is working, I assume that it is Java based?!

No, it is .NET-based. The code is here: https://github.com/kohsuke/winsw/blob/929b87c383ab1ac89afff1d2ed02162be49aaddc/src/Core/WinSWCore/Download.cs . So it just downloads via the GET request using the specified `from` field, no other options. In this ticket I consider extending the flexibility of the referenced class

> The Jenkins server is using the reverse proxy plugin for authentication in combination with Kerberos authentication via mod_aut_kerb. The connection to the master is working but the download fails.

"jnlpJars/slave.jar" is a publicly accessible endpoint, so you should not need authentication. Likely the issue is with the self-signed certificate. Likely you will need to somehow register it in the system

nightman68:
Hi Oleg,

thanks for the quick response!

I configured the Apache to require authentication for all content. I know that it is recommended to leave some content open but we have constraints and this behavior keeps the Apache configuration simple - my POV.

I will try to check what the code gives for opportunities to add an optional authentication and will come back later on.

oleg-nenashev:
Likely it will be done as an enhancement of the existing `download` configuration entry. #195 and #197 are the pending changes

oleg-nenashev:
I assume it is closed in 2.1.0
Other enhancements can be submitted as separate issues

nightman68:
[...]

PR #194, opened 2017-02-24
Issue #126 - Add support of authentication in the download settings

nightman68:
This code adds authentication handling for downloads from  a server
which requires this. [Source](http://stackoverflow.com/questions/5572588/webrequest-getresponse-is-throwing-error-401-unauthorized).

I tested the functionality with a Jenkins instance using the plugin "HTTP Header by reverse proxy" and an Apache 2.4 with Kerberos authentication (mod_auth_kerb 5.4). Another check was to "disable" any authentication that no credentials were required (Apache "Require all granted").

The service was created on Windows server 2008 with bind to a domain account.

### DIFF
--- a/src/Core/WinSWCore/Download.cs
+++ b/src/Core/WinSWCore/Download.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.IO;
 using System.Net;
 using System.Xml;
@@ -23,6 +23,10 @@ namespace winsw
         public void Perform()
         {
             WebRequest req = WebRequest.Create(From);
+            req.UseDefaultCredentials = true;
+            req.PreAuthenticate = true;
+            req.Credentials = CredentialCache.DefaultCredentials;
+
             WebResponse rsp = req.GetResponse();
             FileStream tmpstream = new FileStream(To + ".tmp", FileMode.Create);
             CopyStream(rsp.GetResponseStream(), tmpstream);
--- a/src/Core/WinSWCore/Download.cs
+++ b/src/Core/WinSWCore/Download.cs
@@ -1,6 +1,7 @@
 using System;
 using System.IO;
 using System.Net;
+using System.Text;
 using System.Xml;
 
 namespace winsw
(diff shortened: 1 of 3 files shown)


nightman68:
It's a good point. I'm not a .NET programmer but I understood that this auth. is not supported on un-encrypted connections like HTTP and FTP, [https://msdn.microsoft.com/de-de/library/system.net.credentialcache.defaultcredentials(v=vs.85).aspx](url).

I will check this, also a possible opt-in.

nightman68:
I enhancement is now supporting all kind of Microsoft SSPI authentication and a basic auth. All auth. types has been tested with Apache and the reverse proxy plugin for Jenkins.

Because it's my first C# programming I assume that the handling of the XML arguments is not the perfect one!?

I'm still extending the documentation.

src/Core/WinSWCore/Download.cs
```diff
@@ -11,18 +12,91 @@ namespace winsw
...
+            try
+            {
+                tmpStr = Environment.ExpandEnvironmentVariables(n.Attributes["unsecureAuth"].Value);
+            }
+            catch (Exception)
+            {
+            }
+            UnsecureAuth = tmpStr == "enable" ? true : false;
+
+            if (Auth == AuthType.basic)
```
nightman68: for basic auth. the transfer should be encrypted otherwise the credentials are send in clear text!

oleg-nenashev:
@nightman68 Thanks for the change! From the functionality side it definitely makes sense to integrate it. From the code side I can take care of the refactoring/integration once you are fine with the documentation side

src/Core/WinSWCore/Download.cs
```diff
@@ -11,18 +12,91 @@ namespace winsw
...
+            {
+                if (From.StartsWith("http:") && UnsecureAuth == false)
+                {
+                    throw new Exception("Warning: you're sendig your credentials in clear text to the server. If you really want this you must enable this in the configuration!");
```
oleg-nenashev: typo: "sending"

oleg-nenashev: I'd say that HTTP check is generally not enough to say the auth is "secure". I would expect the HTTPS to be checked explicitly, and maybe the code should also verify the certificate, etc.

Maybe it just needs a proper documentation.

nightman68: This is of course is a point to discuss about .-) 

My intention testing HTTP was, that the user is aware about the insecure communication, see also the documentations (part of the commit). For HTTPS I think the procedure must be as follow:
* the CA of the server certificate must be imported on OS side
* if the CA is in the right place the OS is checking the validity of the certificate/connection

I would say that an additional check on our side is not necessary. Depending on the Win release the connection is blocked when the CA is not present. On Windows 10 for instance EDGE and also the IE do not establish a connection and they also not giving the user the opportunity to import the server certificate on the fly. These behavior is the same for this program - the same libraries used.

What do you think?

I will extend the documentation regarding the CA handling. I will commit it together with the type fix.

oleg-nenashev: Yep, it is not a blocker for the merge.
I'd say the behavior should be just documented in WinSW docs and the config sample. In such case it is possible to just say "Use this feature on your own risk".

nightman68:
Hi Oleg,

I added some more details in the documentation and the sample configuration file. From my side this should be OK now. When you need additional changes please let me know.

Tomorrow afternoon I leave for vacation for 2 1/2 week without a laptop!

Regards
Torsten

oleg-nenashev:
> Tomorrow afternoon I leave for vacation for 2 1/2 week without a laptop!

Have fun!
I will try to integrate the changes during this time.

nightman68:
I will; sounds good.

CU

PR #195, opened 2017-02-26
Introduce the Download#FailOnError option.

oleg-nenashev:
This change allows preventing the executable start if the requested file cannot be downloaded. It also adds logging of download operations to the wrapper log.

- [x] Code
- [x] Test
- [x] Documentation

The approach still needs to be finalized, because `FailOnError` does not propagate errors to the `start` command (as all other similar commands)

CC @nightman68 I just decided to write it in order to provide an example. If you want, I can integrate your auth proposals in the same way.


nightman68:
Hi Oleg,

I updated the pull request. Have a look to it and give me a hint how to proceed.

oleg-nenashev:
@nightman68 FYI will merge it to the master and then refactor and finalize both changes

nightman68:
Sounds good 👍 

Waiting that it is integrated in Jenkins LTS.

PR #197, opened 2017-02-28
´download auth alternative

nightman68:
Hi Oleg,

I created an alternative for the authentication - less parameters need to be handled. The client tries to authenticate via Basic when credentials are specified. In the case of failure a SSPI authentication is initiated which also works for a server without enabled authentication. When this connection fails the program should abort - something else is not OK.

It's your choice what pull request should be merged .-)

### DIFF
--- a/doc/xmlConfigFile.md
+++ b/doc/xmlConfigFile.md
@@ -147,8 +147,33 @@ This feature should be used only for debugging, as some operating systems and ha
 This optional element can be specified multiple times to have the service wrapper retrieve resources from URL and place it locally as a file.
 This operation runs when the service is started, before the application specified by `<executable>` is launched.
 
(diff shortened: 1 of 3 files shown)


nightman68:
Hi Oleg,

I'm back now and we could continue if you have time.

Did you reviewed my proposals?

Regards
Torsten

oleg-nenashev:
Hi, this is still on my plate. Got stuck due to several urgents in the Jenkins project.
The action item is on my side. I need to polish your code a bit and then integrate with my changes

nightman68:
OK, just let me know when somethings is required!

oleg-nenashev:
OK, I am back on it.

My plan is to integrate & release #202 in 2.0.3, and then merge #195 and this PR towards 2.1.0

oleg-nenashev:
After some consideration I decided to go with #194. Will create a separate pull request with polishing

oleg-nenashev:
closing, because #194 and #203 have been integrated

Issue #201, opened 2017-03-31
All executable environment variables are in lowercase starting from WinSW 2.0


oleg-nenashev:
This is a regression I've introduced in WinSW 2.0 by https://github.com/kohsuke/winsw/commit/92367f44ef208bca20c6dfcadb975722fd28763d. I have added `ps.EnvironmentVariables[WinSWSystem.ENVVAR_NAME_SERVICE_ID.ToLower()] = _descriptor.Id;`, which made all variables to be in lowercase in the injected process. Some tools were not ready to that though they are expected to handle Env Vars in the case-insensitive mode in Windows.

Java tools do not do that By default, hence [JENKINS-42744](https://issues.jenkins-ci.org/browse/JENKINS-42744) was reported by Jenkins users.

oleg-nenashev:
Released in 2.0.3

PR #202, opened 2017-03-31
[JENKINS-42744] - Prevent conversion of environment variables to lowercase

oleg-nenashev:
This change intends fixing the regression I've introduced in WinSW 2.0 in https://github.com/kohsuke/winsw/commit/92367f44ef208bca20c6dfcadb975722fd28763d. I have added `ps.EnvironmentVariables[WinSWSystem.ENVVAR_NAME_SERVICE_ID.ToLower()] = _descriptor.Id;`, which made all variables to be in lowercase in the injected process. Some tools were not ready to that though they are expected to handle Env Vars in the case-insensitive mode in Windows. But Java tools do not do that By default, hence [JENKINS-42744](https://issues.jenkins-ci.org/browse/JENKINS-42744) was reported by Jenkins users.

Changes:

- [x] - Fix Issue #201 (JENKINS-42744)
- [x] - Refactor the `Main#StartProcess()` logic to make it testable
- [x] - Add tests for #201 
- [x] - Also add the process termination tests to the `RunawayProcessKiller` to ensure there is no regressions

https://issues.jenkins-ci.org/browse/JENKINS-42744

@reviewbybees

### DIFF
--- a/src/Plugins/RunawayProcessKiller/RunawayProcessKillerExtension.cs
+++ b/src/Plugins/RunawayProcessKiller/RunawayProcessKillerExtension.cs
@@ -118,9 +118,10 @@ namespace winsw.Plugins.RunawayProcessKiller
             // TODO: This method is not ideal since it works only for vars explicitly mentioned in the start info
             // No Windows 10- compatible solution for EnvVars retrieval, see https://blog.gapotchenko.com/eazfuscator.net/reading-environment-variables
             StringDictionary previousProcessEnvVars = proc.StartInfo.EnvironmentVariables;
-            String expectedEnvVarName = WinSWSystem.ENVVAR_NAME_SERVICE_ID.ToLower();
+            String expectedEnvVarName = WinSWSystem.ENVVAR_NAME_SERVICE_ID;
             if (previousProcessEnvVars.ContainsKey(expectedEnvVarName))
             {
+                // StringDictionary is case-insensitive, hence it will fetch variable definitions in any case
                 affiliatedServiceId = previousProcessEnvVars[expectedEnvVarName];
             }
             else if (CheckWinSWEnvironmentVariable)
--- a/src/Plugins/RunawayProcessKiller/RunawayProcessKillerExtension.cs
+++ b/src/Plugins/RunawayProcessKiller/RunawayProcessKillerExtension.cs
@@ -27,6 +27,12 @@ namespace winsw.Plugins.RunawayProcessKiller
         /// </summary>
         public bool StopParentProcessFirst { get; private set; }
 
+        /// <summary>
+        /// If true, the runaway process will be checked for the WinSW environment variable before termination.
+        /// This option is not documented AND not supposed to be used by users.
+        /// </summary>
+        public bool CheckWinSWEnvironmentVariable { get; private set; }
+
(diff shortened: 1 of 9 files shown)


oleg-nenashev:
@jglick I have splitted the chanve to multiple commits for a reason. Here is the actual fix: https://github.com/kohsuke/winsw/pull/202/commits/f81f5d3c574d6b72f08a9325845c25541e6c34ab

oleg-nenashev:
@reviewbybees done

src/Plugins/RunawayProcessKiller/RunawayProcessKillerExtension.cs
```diff
@@ -38,9 +44,12 @@ public RunawayProcessKillerExtension()
             // Default initializer
         }
 
-        public RunawayProcessKillerExtension(String pidfile)
+        public RunawayProcessKillerExtension(String pidfile, int stopTimeoutMs = 5000, bool stopParentFirst = false, bool checkWinSWEnvironmentVariable = true)
         {
             this.Pidfile = pidfile;
+            this.StopTimeout = TimeSpan.FromMilliseconds(5000);
```
oleg-nenashev: This is a bug, but this API is not being used anywhere outside tests. will fix in a follow-up

src/Plugins/RunawayProcessKiller/RunawayProcessKillerExtension.cs
```diff
@@ -38,9 +38,11 @@ public RunawayProcessKillerExtension()
             // Default initializer
         }
 
-        public RunawayProcessKillerExtension(String pidfile)
+        public RunawayProcessKillerExtension(String pidfile, int stopTimeoutMs = 5000, bool stopParentFirst = false)
```
daniel-beck: Unused argument

src/Plugins/RunawayProcessKiller/RunawayProcessKillerExtension.cs
```diff
@@ -113,11 +116,12 @@ public override void OnWrapperStarted()
             else
             {
                 Logger.Warn("The process " + pid + " has no " + expectedEnvVarName + " environment variable defined. " 
-                    + "The process has not been started by this service, hence it won't be terminated.");
+                    + "The process has not been started by WinSW, hence it won't be terminated.");
```
daniel-beck: Isn't it `jenkins.exe` for Jenkins? Is it reasonable to expect users to know what WinSW is where this message will show up?

oleg-nenashev: WinSW is not for Jenkins. It is a standalone project, which is being widely used outside. But I agree the message is not ideal. The intention was to say that "The process has not been created by any of winsw-powered services you may have on this machine", but maybe it is better to revert the change

src/Plugins/RunawayProcessKiller/RunawayProcessKillerExtension.cs
oleg-nenashev: Yep, commented in the PR. Will fix in the follow-up though it is harmless in the current version (WinSW API is for internal use only)

PR #203, opened 2017-04-11
Finalize the Download Command changes

oleg-nenashev:
This PR finalizes #126, which includes #194 from @nightman68 and #195 from me.

- [x] Refactor the code
- [x] Switch `unsecureAuth` to the boolean flag, update docs
- [x] Make `auth` field case-insensitive
- [x] Add unit tests for the authentication functionality
- [x] Update configuration samples

### DIFF
--- a/examples/sample-allOptions.xml
+++ b/examples/sample-allOptions.xml
@@ -251,18 +251,20 @@ SECTION: Environment setup
   -->
   <!--
   <download from="http://www.google.com/" to="%BASE%\index.html" />
-  <download from="http://www.nosuchhostexists.com/" to="%BASE%\dummy.html" />
+  
+  Download and fail the service startup on Error:
+  <download from="http://www.nosuchhostexists.com/" to="%BASE%\dummy.html" failOnError="true"/>
 
   An example for unsecure Basic authentication because the connection is not encrypted:
   <download from="http://example.com/some.dat" to="%BASE%\some.dat"
-            auth="basic" unsecureAuth=“enabled”
+            auth="basic" unsecureAuth=“true”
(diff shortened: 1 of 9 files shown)


oleg-nenashev:
@nightman68 Ready for review. CC @reviewbybees just in case

examples/sample-allOptions.xml
```diff
@@ -251,18 +251,20 @@ SECTION: Environment setup
...
-            auth="basic" unsecureAuth=“enabled”
+            auth="basic" unsecureAuth=“true”
             username="aUser" password=“aPassw0rd" />
 
   Secure Basic authentication via HTTPS:
   <download from="https://example.com/some.dat" to="%BASE%\some.dat"
             auth="basic" username="aUser" password="aPassw0rd" />
 
-  Secure authentication when the target server and the client are members of domain:
+  Secure authentication when the target server and the client are members of the same domain:
```
nightman68: The membership in "the same" domain is not normally not the constraint. When a forest trust is defined correctly it is also working.

oleg-nenashev: Right. "of domain" does not seem the best wording as well. Could it be there a better clarification text?

nightman68: Right .-) My proposal:
Secure authentication when the target server and the client are members of the same domain or the server domain and the client domain belong to the same forest with a trust.

oleg-nenashev: updated

nightman68: 👍

oleg-nenashev:
Seems it is ready to go

PR #205, opened 2017-04-14
#183 - Add core logic for the DelayedStart option support

### LINKED ISSUE #183: Add delayed start support
Unless I'm missing something, there is no way to specify 'Automatic (Delayed Start)' as Startup type. This is the `DelayedAutoStart` boolean property of `ServiceInstaller` in .NET.

oleg-nenashev:
Fixes #183 . The only problem is that MMC and Service are too smart and cache the option internally (see [this article](https://social.msdn.microsoft.com/Forums/en-US/e5dbfe02-5125-4833-8e6b-3aa3d8518fc8/delayed-start-registry-setting?forum=biztalkgeneral)), so the setting does not get displayed there until the system restart.

- [x] - Internal installation logic
- [x] - Documentation update
- [x] - Samples update
- [x] - Functional tests

CC @mtman

### DIFF
--- a/src/Core/ServiceWrapper/Main.cs
+++ b/src/Core/ServiceWrapper/Main.cs
@@ -605,6 +605,13 @@ namespace winsw
                     Registry.LocalMachine.OpenSubKey("System").OpenSubKey("CurrentControlSet").OpenSubKey("Services")
                         .OpenSubKey(d.Id, true).SetValue("Description", d.Description);
 
+                    if (d.StartMode == StartMode.Automatic && d.DelayedStart)
+                    {
+                        //TODO: replace by a better API after migrating to .NET 4
+                        Registry.LocalMachine.OpenSubKey("System").OpenSubKey("CurrentControlSet").OpenSubKey("Services")
+                            .OpenSubKey(d.Id, true).SetValue("DelayedAutoStart", 1);
+                    }
+
                     var actions = d.FailureActions;
                     if (actions.Count > 0)
                     {// set the failure actions
--- a/src/Core/ServiceWrapper/Main.cs
+++ b/src/Core/ServiceWrapper/Main.cs
@@ -605,21 +605,25 @@ namespace winsw
                     Registry.LocalMachine.OpenSubKey("System").OpenSubKey("CurrentControlSet").OpenSubKey("Services")
                         .OpenSubKey(d.Id, true).SetValue("Description", d.Description);
 
-                    if (d.StartMode == StartMode.Automatic && d.DelayedStart)
-                    {
-                        //TODO: replace by a better API after migrating to .NET 4
-                        Registry.LocalMachine.OpenSubKey("System").OpenSubKey("CurrentControlSet").OpenSubKey("Services")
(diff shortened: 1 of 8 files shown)


oleg-nenashev:
@reviewbybees [JENKINS-29496](https://issues.jenkins-ci.org/browse/JENKINS-29496)

oleg-nenashev:
CC @jtnord

src/Core/ServiceWrapper/Main.cs
```diff
@@ -605,6 +605,13 @@ public static void Run(string[] _args, ServiceDescriptor descriptor = null)
                     Registry.LocalMachine.OpenSubKey("System").OpenSubKey("CurrentControlSet").OpenSubKey("Services")
                         .OpenSubKey(d.Id, true).SetValue("Description", d.Description);
 
+                    if (d.StartMode == StartMode.Automatic && d.DelayedStart)
+                    {
+                        //TODO: replace by a better API after migrating to .NET 4
+                        Registry.LocalMachine.OpenSubKey("System").OpenSubKey("CurrentControlSet").OpenSubKey("Services")
```
jtnord: This is entirely the wrong way to do this. You should be using the SCM api for all of the service registry - hence the fact the SCM will not show the updated options until reboot.

oleg-nenashev:
@jtnord Agreed. First time I was unable to find a proper API call in the documentation, should be fixed now. 
I have also created #208 for the Service description field

src/Core/WinSWCore/Native/Advapi32.cs
```diff
@@ -79,6 +79,26 @@ public void ChangeConfig(TimeSpan failureResetPeriod, List<SC_ACTION> actions)
...
+        public void SetDelayedAutoStart(bool enabled)
+        {
+            SERVICE_DELAYED_AUTO_START settings = new SERVICE_DELAYED_AUTO_START
+            {
+                fDelayedAutostart = enabled
+            };
+
+            if (!Advapi32.ChangeServiceConfig2(Handle, SERVICE_CONFIG_INFOLEVEL.SERVICE_CONFIG_DELAYED_AUTO_START_INFO, ref settings))
+            {
+                throw new Exception("Failed to change the DelayedAutoStart setting", new Win32Exception());
```
jtnord: should this not have a `GetLastError()` call so it is recorded - or does the above call not set it?

oleg-nenashev:
> Am left wondering if delayed start is correct, as opposed to setting correct service dependencies, but should work all the same (just with more delay)

WinSW allows defining service dependencies. IMHO extra feature should not hurt here, especially since it's hard to define which services may be actually required by Jenkins jobs, for example.

src/Core/WinSWCore/Native/Advapi32.cs
oleg-nenashev: Empty constructor captures in automatically: https://msdn.microsoft.com/en-us/library/asy2azk0(v=vs.110).aspx

oleg-nenashev:
@reviewbybees done

PR #207, opened 2017-04-15
Update xmlConfigFile.md

nightman68:
The URL for the CA import was wrong.


oleg-nenashev:
Yes, the new one is definitely better

nightman68:
Was my mistake, you're welcome!

Issue #148, opened 2016-12-04
Update log4net packages to 2.0.latest


oleg-nenashev:
It would effectively update log4net from `1.2.13` to `1.2.15`. There is no strongly related fixes, but it would be great to have the latest version.

See https://logging.apache.org/log4net/release/release-notes.html

NextTurn:
@oleg-nenashev This is done.

oleg-nenashev:
Yes, eons ago. https://github.com/kohsuke/winsw/commit/a8475e5dfa2e56c43ab7dc6dca07cb7d2cea4826

Issue #87, opened 2015-02-02
Migrate STDERR/STDOUT logs to log4net


oleg-nenashev:
WinSW 1.x uses a home-made implementation of Rolling File Log Adapter.
Usage of a standard library could be useful to decrease the code complexity.

It's a follow-up to #73

oleg-nenashev:
Excluding from the 2.0 milestone for a while.
Event/System logs have been imgratted as a part of #69

aminjam:
@oleg-nenashev Thanks for the context. We would like to help implement this feature. If you have the full scope laid out we would be happy to send a PR for this feature. Let us know if you have specific implementation detail in mind.

oleg-nenashev:
The feature itself is pretty easy to implement (more easier than to explain how it should be implemented), but I will appreciate testing and code reviews once it is ready. Would you be available on the next week to test it?

aminjam:
Yes. Let us know whenever it's ready.

oleg-nenashev:
OK, added more detailed description to the issue

PR #212, opened 2017-04-23
Issue #148 - Update log4net to 2.0.8

oleg-nenashev:
It effectively bumps the bundled log4net version from 1.2.13 to 2.0.8 (log4net NuGet package used to have different versions). According to the changelogs, there is no expected compatibility issues: 

log4net changelog: https://logging.apache.org/log4net/release/release-notes.html

Addresses #148


oleg-nenashev:
Works fine according to manual tests. Merging to unblock #87

Issue #206, opened 2017-04-14
Status command text contains log output


Digital-Chaos:
$ winsw status
The above command now emits INFO logging text in addition to the expected service status string due to ConsoleLogger being enabled if in "cli mode" (which seems to be if you have args > 0)??? There also seems no way to disable this logging from the configuration?

oleg-nenashev:
Not for now. https://github.com/kohsuke/winsw/issues/158 should add support of it.

BTW I doubt that "status" command will ever guarantee to provide a short output since there is some long-term plan to have more diagnostics there. What is your use-case? Is it for monitoring?

Digital-Chaos:
The current docs on the front page say that the status command will return one line which indicates the status of the service with values of "NonExistant", "Started", or "Stopped", which is no longer true. 

Our use case is that we have automated deployment of a Spring Boot app in which we have an initial powershell step which checks the state of the service (utilizing the result of the winsw status command) and then stops it if running. There is also a final step after deployment to check for the existance of the service (again using winsw status) and install it if necessary and then start the service. 

Why change the exisitng "status" command to add extra diagnostics output and potentially cause breakage for existing users when you could add an additional command such as "winsw diagnotics" or something similiar and avoid breaking anything?

oleg-nenashev:
If it is documented, I agree it should behave in the described way. Adding it to the 2.2 scope

Digital-Chaos:
Much appreciated, and thank you for your time and effort maintaining this project! :-)

oleg-nenashev:
Released in 2.1.1.
Sorry for the delay. I should spend more time on establishing Continuous Delivery in my projects :(

oleg-nenashev:
Looks to be completed, the command works well on the current version. Please feel free to reopen if I missed anything

PR #214, opened 2017-04-25
[Issue #206] - Prevent printing of logs in the status command

oleg-nenashev:
Addresses #206, regression in 2.0 since it breaks the documented behavior. This is rather a workaround to prevent printing of log messages in CLI with the default logging level. 

* [x] Reduce CLI logging verbosity to `Info` 
* [x] Move the messages polluting the output to `Debug`. Anyway, these logging levels seem to be reasonable. 
* [x] Add `Info` logging for start/stop/install/restart commands to have a welcome message in CLI

I do not consider it as final fix. Ideally WinSW startup logic should determine commands like `status` and disable console log appender for them at all. It requires extra changes in the code.

@reviewbybees @Digital-Chaos

### DIFF
--- a/src/Core/ServiceWrapper/Main.cs
+++ b/src/Core/ServiceWrapper/Main.cs
@@ -463,7 +463,7 @@ namespace winsw
             try
             {
                 Run(args);
-                Log.Info("Completed. Exit code is 0");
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
@reviewbybees done

Issue #218, opened 2017-05-10
writing to stdout from a stopexecutable will eventually cause the process to hang


pnikonowicz:
This happens because the stop executable's stdout stream is redirected to the parent process but is never drained, thus eventually causing the stopexecutable's stdout writer to hang. 

I recommend either setting these flags to false here:
https://github.com/kohsuke/winsw/blob/0da384180419ae5557685736e1e0e5fdab17eeff/src/Core/WinSWCore/Util/ProcessHelper.cs#L141

Or actually consuming the processes stream and piping it to a log file. 

For more info on how to redirect stdout from a process, see here:
https://msdn.microsoft.com/en-us/library/system.diagnostics.processstartinfo.redirectstandardoutput(v=vs.110).aspx

oleg-nenashev:
> Or actually consuming the processes stream and piping it to a log file

This one sounds like a plan

oleg-nenashev:
The change also impacts the start (common) executable. According to the brief testing it is not correct in that case because the logs are being drained correctly there to logfiles. I will revisit the implementation, hopefully within 1-2 days

oleg-nenashev:
The fix has been released in 2.1.1
Once I get #213 integrated, I will setup forwarding of the output to the wrapper log

PR #213, opened 2017-04-23
[Issue #87] - Make logging fully configurable via log4net

oleg-nenashev:
Addresses #87 and several other feature requests regarding logging subsystem (#182, #39, #61). Now everybody will be able to configure whatever he wants using the `log4net` logging mode.

Current implementation supports two modes:

1. Redirect STDERR/STDOUT to log4net (Info and Error logging levels)
2. Redirect STDERR/STDOUT to log4net **AND** configure all loggers (CLI, Wrapper log, Event log, STDOUT/STDERR) over a standard [log4net configuration file](https://logging.apache.org/log4net/release/manual/configuration.html)

Mode 1: Redirect STDERR/STDOUT

```xml
  <log mode="log4net"/>
```

Mode 2: Full configuration via config file

```xml
  <log mode="log4net">
	<configPath>%BASE%\log4net.xml</configPath>
  </log>
```

TODOs:
- [x] - First prototype
- [ ] - rework PoC to have two separate logging mode names
- [ ] - Get feedback from users and requesters
- [ ] - Polish/refactor the implementation
- [ ] - Documentation
- [ ] - Functional tests

Test build will be available from AppVeyor. CC @aminjam , @roederja2 , @maurokade, @Randy1967, @pvandervelde

PR #224, opened 2017-06-08
Issue #218 - ProcessHelper#StartProcessAndCallbackForExit() should redirect STDOUT/STDERR when LogHandler is defined

oleg-nenashev:
It restores logging of executables, which has been broken in https://github.com/kohsuke/winsw/pull/220.
Not a regression, because the change has not been released yet. 

It is not a full fix, I still need to determine why it causes a preliminary service exit in the case of the infinite ping executable. 

CC @pnikonowicz


oleg-nenashev:
Works fine for me now.
A better fix depends on the new appender logic in #213

oleg-nenashev:
Tested it locally, works fine for me

PR #227, opened 2017-06-27
Noticed a small 'error' and fixed it

Darsstar:
Textbook example of what will eventually happen if you copy-paste enough snippets?


oleg-nenashev:
Yeah, it happens sometimes :(

oleg-nenashev:
Fortunately the values are same anyway, hence no user-visible impact

Darsstar:
> Fortunately the values are same anyway, hence no user-visible impact

Hence the 'error' :)

PR #228, opened 2017-06-28
Start using stopTimeoutMs

Darsstar:


oleg-nenashev:
Will release it by Monday

Issue #237, opened 2017-07-28
RunawayProcessKiller kill wrong process


Thomehack:
After Windows reboot, the WinSW kill another process with the same pid was jenkins has before shutdown.

Simple test:
- stop service
- start a cmd.exe
- write pid of cmd.exe to jenkins_agent.pid
- start service
=> WinSW kill the cmd.exe

oleg-nenashev:
I have a check environment variable in the code: https://github.com/kohsuke/winsw/blob/master/src/Plugins/RunawayProcessKiller/RunawayProcessKillerExtension.cs#L34 . It is true by default in the initializer ([code](https://github.com/kohsuke/winsw/blob/8d7f81f213a5809f8ce8aac2d295ca4c39849adc/src/Plugins/RunawayProcessKiller/RunawayProcessKillerExtension.cs#L64-L65)), so it should also check the environment variable.

Could you please provide your Process Killer configuration?

Thomehack:
Do you mean the checkWinSWEnvironmentVariable node?
It is not set i my configuration.

=> Especially, the jenkins agent service run as the logged on user.

Can you create a binary (for a test) where printing the propertys CheckWinSWEnvironmentVariable, ServiceId and the variable affiliatedServiceId before you killing a process.

??? I cant build it the project. log4net is missing

oleg-nenashev:
> Can you create a binary (for a test) where printing the propertys CheckWinSWEnvironmentVariable, ServiceId and the variable affiliatedServiceId before you killing a process.

I can add a patch to main binary for it (e.g. on the DEBUG log level). But I can work on it only on Sunday - my Windows dev environment is 600km away.

> ??? I cant build it the project. log4net is missing

It comes from NuGet. MSBuild should automatically download it if something is not messed up. Could you please provide build log?

oleg-nenashev:
If the project does not build locally due to whatever reason, just create a pull request with your patch. In such case you will be able to download the binary from Appveyor if the build passes

Thomehack:
<blockquote>
I can add a patch to main binary for it (e.g. on the DEBUG log level). But I can work on it only on Sunday - my Windows dev environment is 600km away.
</blockquote>

That's nice of you.
Sunday is not a problem, my next working day is wednesday.

oleg-nenashev:
> Sunday is not a problem, my next working day is wednesday. 

Hehe, same for me. Does not apply to open-source activities though ;)

[...]

PR #239, opened 2017-08-01
Issue #237 - Improve diagnostics of the Runaway Process Killer when it kills the process

oleg-nenashev:
See #237. The change adds some extra diagnostics, which may help to determine if we kill the right process. 

For more hardening, I could also add the executable name check. After such change WinSW will ignore Runaway processes if the executable name changes. This is not a real case IMHO.

@Thomehack @reviewbybees


oleg-nenashev:
Printing of the process is safe from what I see.
Merging as is, will release along with #240.

Issue #233, opened 2017-07-13
Issue with DefferedFileOperations


vladotod:
I am using WinSW 1.18 on Windows 10. My application uses DefferedFileOperations for auto-update. There is strange behaviour when trying to overwrite target file but source file not found. In this case WinSW will delete target file. 

**How to reproduce this issue**

- Create empty folder: D:\winsw\source 
- Create target file: D:\winsw\target\test.txt
- Create myapp.copies file with one line: `D:\winsw\source\test.txt>D:\winsw\target\test.txt`
- Copy myapp.copies to root dir of myapp application and restart myapp Windows service.

After restart of myapp, WinSW will delete myapp.copies (it is expected) but also will delete target file `D:\winsw\target\test.txt` 

Windows EventViewer contains these messages:

1. Handling copy: D:\winsw\source\test.txt>D:\winsw\target\test.txt
2. Failed to copy :D:\winsw\source\test.txt to D:\winsw\target\test.txt because Could not find file 'D:\winsw\source\test.txt'.

oleg-nenashev:
This is how it behaves now: https://github.com/kohsuke/winsw/blob/a2c2a9d40550565fbbd3eb3cfa0175898f1a588c/src/Core/ServiceWrapper/Main.cs#L123-L134
1. Delete the target file
2. Copy the source file

The first operation happens even if the source file does not exist. I could add a check of the file before execution, so the original file will be retained in such case.

vladotod:
Although this behaviour could be used for deleting files it isn't so intuitive. I also think that target file should be retained if source file doesn't exist.

oleg-nenashev:
Will do it on the weekend

NextTurn:
Fixed in #395.

PR #240, opened 2017-08-02
Issue #233 - Check source existence before moving files

oleg-nenashev:
Addresses #233 

- [x] - Check the source existence before copying
- [ ] - Use `File.Replace` to make operation atomic and to keep backups

CC @vladotod, you can use the current state of the PR as a temporary fix


NextTurn:
Superseded by #395?

oleg-nenashev:
I think so

PR #254, opened 2017-10-30
pass ServiceDescriptor when creating wrapperservice

senarclens:
allow to create configured WrapperService instances


oleg-nenashev:
Will release WinSW this week

PR #277, opened 2018-05-31
Fix environment variable setting in RunawayProcessKiller test

corebonts:
This way the check can be reenabled as well.


oleg-nenashev:
Oops, thanks!

PR #281, opened 2018-06-22
Update installation.md

ricardopolo:
Fixing wrong order in step list


oleg-nenashev:
Thanks @ricardopolo !

PR #286, opened 2018-07-25
maven: also deploy binary targeting .NET 4.0

dmngb:


oleg-nenashev:
Would it be possible to keep `WinSW.NET4.exe` as the name?
Generally deployment to Maven is rather deprecated, but I do not mind adding it there

dmngb:
> Would it be possible to keep WinSW.NET4.exe as the name?

You mean using winsw.net4.exe instead of winsw4.exe when using attachArtifact  ?
This would not change much: this is only a temporary name during the build.

The final names in the maven repository will be in any case something like winsw-2.1.3-bin.exe  (.NET 2) or winsw-2.1.3-net4.exe (.NET 4).

oleg-nenashev:
Thanks @dmngb ! Next released will be shipped with this artifact

dmngb:
Thanks @oleg-nenashev for the super-fast commit !

oleg-nenashev:
OK, it is included into 2.2.0. I will make sure to release versions more often going forward

PR #318, opened 2019-05-20
Fix spelling in samples

pram:
Fixing spelling


oleg-nenashev:
Good news, typos have test coverage 🤦‍♂ 

```
1) Test Failure : winswTests.Configuration.ExamplesTest.allOptionsConfigShouldDeclareDefaults
     String lengths are both 61. Strings differ at index 28.
  Expected: "This service is a service cratead from a sample configuration"
  But was:  "This service is a service created from a sample configuration"
  ---------------------------------------^
```

oleg-nenashev:
Thanks @pram !

PR #321, opened 2019-05-28
Add a new `hidewindow` option to suppress windows popup on legacy platforms

jeacott1:
adds a `<hidewindow>true</hidewindow>` config switch
so that launching bat files on legacy machines
running Interactive Services Detection doesn't balk with horrid warnings.
default behaviour remains as before.


oleg-nenashev:
Sorry, I was off for too long in this repo

PR #330, opened 2019-08-13
Fix duplicate startup logging

ryanrupp:


ryanrupp:
Noticed this in the startup log e.g.:

```
2019-08-13 14:02:36,807 INFO  - Starting java -Xmx128m -jar "my.jar"
2019-08-13 14:02:36,823 INFO  - Starting java -Xmx128m -jar "my.jar"
```

Not sure if the second set of logging is meant to be for mentioning that extensions were loaded or something or was maybe just the result of a merge conflict being resolved with duplicate lines. Duplicate was introduced here - https://github.com/kohsuke/winsw/commit/2546168ed395ab992ca357dddd02d861cf885cf5#diff-aa1cd4852033c3fac020e2c2d6c29637R267

PR #341, opened 2019-10-17
Add stopwait command - same as stop, but waits until the service is actually stopped:

new-mikha:
The command is useful in uninstall scenarios, when the service needs to be stopped first.

It's better than `net stop <serviceId>` because the new `stopwait` command doesn't need to specify the actual service ID. I.e. it allows to keep the service ID in one place rather than two (config XML and uninstall script)

### DIFF
--- a/src/Core/ServiceWrapper/Main.cs
+++ b/src/Core/ServiceWrapper/Main.cs
@@ -607,6 +607,10 @@ namespace winsw
                     Stop();
                     return;
 
+                case "stopwait":
+                    StopWait();
+                    return;
+
                 case "restart":
                     Restart();
                     return;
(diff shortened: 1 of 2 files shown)


new-mikha:
Hello @oleg-nenashev , could you please have a look at this feature? 
Let me know please what's needed from my side.
Thanks!

new-mikha:
> Sorry, I have missed the pull request. I should spend more time on this repository :(
> The feature looks great, +1 for adding it. Could you please also add it to https://github.com/kohsuke/winsw#usage ?

Thanks!! I've just updated the README too

oleg-nenashev:
OK, resolving conflicts in GitHub breaks encoding. My apologies, forgot about this PR

src/Core/ServiceWrapper/Main.cs
```diff
@@ -815,6 +819,25 @@ void Stop()
...
+                Log.Info("Stopping the service with id '" + descriptor.Id + "'");
+                if (s is null)
+                    ThrowNoSuchService();
+
+                if (s.Started)
+                    s.StopService();
+
+                while (s != null && s.Started)
+                {
+                    Log.Info("Waiting the service to stop...");
```
oleg-nenashev, line 833: Should we really print it on every cycle in the loop?

new-mikha: @oleg-nenashev Well, I think it's useful that the wrapper shows that it's not completely stuck, but it's up to you

NextTurn: I agree that we print something to indicate that the program isn't stuck. Later we can do something interesting such as drawing a progress bar in the console.

oleg-nenashev:
Thanks for this new feature @new-mikha ! And sorry for the merge delay, we should have more maintenance bandwidth going forward

PR #352, opened 2020-01-13
Migrate to SDK-style projects + Require Visual Studio Community 2019 to build the project

NextTurn:
Closes #149 
Closes #261 

A prerequisite to add more TFMs. #337, #344 

Note: This PR does not intend to make any functionality changes.

PR #353, opened 2020-01-13
Code Style and Formatting cleanup

NextTurn:
Enforces most analyzer rules other than naming ones.

Based on #352.


oleg-nenashev:
Unfortunately I have no Admin permission inside this repo, so I cannot tweak GitHub settings to get a merge with the master button. Will see whether I can work it around by changing branches

src/Core/ServiceWrapper/Main.cs
```diff
@@ -356,41 +348,40 @@ private void WaitForProcessToExit(Process processoWait)
...
             }
 
-//            WriteEvent("WaitForProcessToExit [finished]");
+            // WriteEvent("WaitForProcessToExit [finished]");
         }
 
         private void SignalShutdownPending()
         {
             int effectiveWaitHint;
-            if (_descriptor.WaitHint.TotalMilliseconds > Int32.MaxValue)
```
oleg-nenashev: IMHO explicit type is better for readability, especially for those ones who are not familiar with .NET type standards

NextTurn:
> so I cannot tweak GitHub settings to get a merge with the master button.

What's the branch policy in this repo? And does the problem still exist? I see you eventually merged this into master.

src/Core/ServiceWrapper/Main.cs
NextTurn: There are some additional considerations:

https://blog.paranoidcoding.com/2019/04/08/string-vs-String-is-not-about-style.html

oleg-nenashev:
I have permissions to merge pull requests, manage branches, etc. (Write permission). But I have no access to settings of this repo, so my powers here are limited. I can alwats reach out to @kohsuke if needed tho.

Later I would like to move the repo to a Foundation or to a new GitHub org so that permissions could be granted to more contributors there. I consider Jenkins or a .NET foundation though I am not sure WinSW would be accepted to the latter one. Except being implemented in .NET, it does not have much to offer specifically to .NET projects at the moment

NextTurn:
> Later I would like to move the repo to a Foundation or to a new GitHub org so that permissions could be granted to more contributors there.

Oh really. Did you count how many contributors you can grant permissions to? :)

PR #354, opened 2020-01-14
Add new target frameworks: .NET 4.5, .NET Core win-x64, .NET Core win-x86 + Fix archiving of old logs in the `roll-by-size-time` mode

### LINKED ISSUE #337: Enable TLS 1.1 /1.2 support on Windows versions which support it
Hi we would like to download a jar file from a https secured server. However, the server does not support TLS 1.0, only TLS 1.1 upwards.

Since WINSW does not support TLS 1.1, 1.2 and 1.3 with .Net4.0, we cannot realize our project. 

My question: would you provide a version that supports newer TLS versions (i.e. with .Net 4.5)? That would help us incredibly.

### LINKED ISSUE #344: Executable needed
Hi,
I need an executable for .Net Core

NextTurn:
Closes #337 
Closes #344 
Based on #353 

Is it a hard requirement that WinSW must be published as a single executable? It is possible to do that in .NET Core but it won't bring you any benefits.

This version is not ready for release.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,0 +1,7 @@
+﻿<Project>
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net45' or '$(TargetFramework)' == 'netcoreapp3.1'">
+    <DefineConstants>VNEXT</DefineConstants>
+  </PropertyGroup>
+
+</Project>
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,7 @@ skip_branch_with_pr: true
 
 # Project configuration
 image: Visual Studio 2019
-platform: Any CPU
+# platform: Any CPU
 configuration: Release
 version: 2.0.{build}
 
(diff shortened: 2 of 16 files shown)


oleg-nenashev:
> Is it a hard requirement that WinSW must be published as a single executable? It is possible to do that in .NET Core but it won't bring you any benefits.

Well, we could reconsider the packaging scheme. Single executable packaging is used in downstream projects like Jenkins (https://github.com/jenkinsci/windows-slave-installer-module, https://github.com/jenkinsci/winsw-maven-packaging), but it is something what could be changed. But indeed single-exec is handy for many use-cases

NextTurn:
The produced executable is indeed large, which is about 35MB on my machine.

NextTurn:
It seems that Nuget on CI always tried to find packages from your source and failed, which could even cause build failures if retries count exceeds the limit.

https://ci.appveyor.com/project/oleg-nenashev/winsw/builds/30107995

NextTurn:
https://github.com/kohsuke/winsw/commit/eebc49d295a912ba1d623b8e59c35aa23099dad9

The zip library introduced 2 years ago wasn't merged into the executables at all. I don't see any crash reports, so I think no one actually used those features.

oleg-nenashev:
@NextTurn From your point of view, what is left here? Looks like "This version is not ready for release." is obsolete. 4.5 build passes smoke testsl on my machine

NextTurn:
OK, this is ready for merge. But it's not well prepared to be published in new versions.

.NET 4.5 was an arbitrary choice. Now I'd like to re-target it to .NET 4.6.1 for v.next to catch up with .NET Standard 2.0 support, since it's the lowest version that Nuget considers as supporting .NET Standard 2.0.

https://docs.microsoft.com/dotnet/standard/net-standard#net-implementation-support

oleg-nenashev:
OK, there is no hurry to release it. It can wait for few weeks in 'master', can always release from a branch if there is something urgent. I agree with switching to 4.6.1 before the release

PR #357, opened 2020-01-15
Optimize DynamicProxy

NextTurn:
Based on #356 

This PR made large changes to two "public" classes. I'm assuming this is used as a console program so API changes are OK?

### DIFF
--- a/src/Core/WinSWCore/DynamicProxy.cs
+++ b/src/Core/WinSWCore/DynamicProxy.cs
@@ -19,196 +19,63 @@ namespace DynamicProxy
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


NextTurn:
Consider evaluating the necessity of synchronization, which is blocking further optimization.

oleg-nenashev:
> This PR made large changes to two "public" classes. I'm assuming this is used as a console program so API changes are OK?

Yes, there is no binary compatibility commitment. We can also consider cutting a new major release taking all other changes

NextTurn:
Ping @oleg-nenashev

PR #359, opened 2020-01-16
Fix RunawayProcessKiller

### LINKED ISSUE #237: RunawayProcessKiller kill wrong process

NextTurn:
Fixes #237 
Based on #356 

I'll see if I can eliminate the duplication another day.

### DIFF
--- a/src/Plugins/RunawayProcessKiller/NativeMethods.cs
+++ b/src/Plugins/RunawayProcessKiller/NativeMethods.cs
@@ -1,0 +1,180 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace winsw.Plugins.RunawayProcessKiller
+{
+    public partial class RunawayProcessKillerExtension
+    {
+        internal static class NativeMethods
+        {
+            private const string Kernel32 = "kernel32.dll";
+            private const string NTDll = "ntdll.dll";
+
+            [DllImport(Kernel32)]
+            internal static extern int IsWow64Process(IntPtr hProcess, out int Wow64Process);
+
+            [DllImport(NTDll)]
+            internal static extern int NtQueryInformationProcess(
+                IntPtr ProcessHandle,
+                PROCESSINFOCLASS ProcessInformationClass,
+                out PROCESS_BASIC_INFORMATION32 ProcessInformation,
+                int ProcessInformationLength,
+                IntPtr ReturnLength = default);
+
+            [DllImport(NTDll)]
+            internal static extern int NtQueryInformationProcess(
+                IntPtr ProcessHandle,
+                PROCESSINFOCLASS ProcessInformationClass,
+                out PROCESS_BASIC_INFORMATION64 ProcessInformation,
+                int ProcessInformationLength,
+                IntPtr ReturnLength = default);
+
+            [DllImport(NTDll)]
+            internal static extern unsafe int NtReadVirtualMemory(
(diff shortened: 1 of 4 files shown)


oleg-nenashev:
For the record there is https://github.com/kohsuke/winp which could be used for the same purpose though it would mean using JNI interfaces (or repackaging the lib). Using native API looks fine, but we need to define and document which versions of WinAPI will be supported going forward.

NextTurn:
> For the record there is https://github.com/kohsuke/winp which could be used for the same purpose

How does it work?

oleg-nenashev:
> How does it work?

Well, basically it does the same as what you do in the native code (amd64 or x86 DLLs), but with some magic around WinAPI versions since it still supports WinXP which does not offer WoW64 methods. Taking all the issues with the lib, I am not sure I recommend it, but just in case.

I do not mind dropping old Windows platforms in this PR, but we need to determine what will be the new system requirements

NextTurn:
I **think** a .NET executable will only be run as a 64-bit program on 64-bit systems as long as it doesn't have the `IMAGE_FILE_32BIT_MACHINE` flag (of course we don't define this). The WoW64 APIs are only useful for the .NET Core win-x86 target which obliviously doesn't support Windows XP.

oleg-nenashev:
> I think a .NET executable will only be run as a 64-bit program on 64-bit systems as long as it doesn't have the IMAGE_FILE_32BIT_MACHINE flag (of course we don't define this).

It was definitely running well for the .NET 2 packaging 5 years ago, but I admit I have not used 32bit systems since that, and I am not sure whether it would still run with tool updates we had

NextTurn:
> but I admit I have not used 32bit systems since that

I don't follow. There is nothing to do with 32-bit systems as we don't need WoW64 methods on 32-bit systems.

oleg-nenashev:
Sorry, I misread "I think a .NET executable will only be run as a 64-bit program on 64-bit systems "

oleg-nenashev:
@NextTurn Looks like this PR caused a random failure in the Master branch:

```

C:\projects\winsw\src\Test\winswTests>set  1>C:\Users\appveyor\AppData\Local\Temp\1\winswTests_ysdgyfpf.xcn\env.txt 
C:\projects\winsw\src\Test\winswTests>pause
Press any key to continue . . . Test: Killing runaway process with ID=1052
  X ShouldKillTheSpawnedProcess [331ms]
  Error Message:
     Process System.Diagnostics.Process (cmd) should have been terminated by RunawayProcessKiller
  Expected: True
  But was:  False
  Stack Trace:
     at winswTests.Extensions.RunawayProcessKillerExtensionTest.ShouldKillTheSpawnedProcess() in C:\projects\winsw\src\Test\winswTests\Extensions\RunawayProcessKillerTest.cs:line 106
  Standard Output Messages:
 Created the temporary directory: C:\Users\appveyor\AppData\Local\Temp\1\winswTests_ysdgyfpf.xcn
  Standard Error Messages:
 Test: Killing runaway process with ID=1052

```

https://ci.appveyor.com/project/oleg-nenashev/winsw/builds/30498000

PR #366, opened 2020-01-20
Migrate from `System.Management` to MMI

NextTurn:
Conflicts with #357 

*System.Management* has been deprecated. There are such notes everywhere in the docs.

> **System.Management** was the original .NET namespace used to access WMI; however, the APIs in this namespace generally are slower and do not scale as well relative to their more modern **Microsoft.Management.Infrastructure** counterparts.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -4,4 +4,8 @@
     <DefineConstants>VNEXT</DefineConstants>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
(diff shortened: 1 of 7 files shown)


NextTurn:
MMI carries native libraries, so I'll probably change to only use it on .NET Core.

NextTurn:
> I would recommend to change CIM to a more self-descriptive name

Like `FEATURE_CIM`/`HAVE_CIM`/`WITH_CIM`?

oleg-nenashev:
Yes. Maybe `CIM_ENABLED` and a link to https://docs.microsoft.com/en-us/windows/win32/wmisdk/common-information-model in comments

Issue #68, opened 2015-01-30
Provide XSD schema for XML configuration files


oleg-nenashev:
Use-cases:
* Validate common options and settings
* Assist the configuration of complex cases like Extension points ( #42 )

NextTurn:
I don't think it's possible to design such a schema where some elements are unique, some are repeatable, and all of them can appear in any order at the same level.

Either non-unique elements shall be changed to child elements of grouping ones, or all elements shall be grouped (like *.csproj*).

oleg-nenashev:
XSD actually supports all of that. I just do not hurry with that, would be better to just move to YAML. But YAML/JSON Schemas are really weak being compared to XSD

NextTurn:
That would be very complicated.

See this example with only two unique elements: https://stackoverflow.com/a/12012599

NextTurn:
It's only possible with an exception that the same repeatable elements are placed together.

Yes

```xml
<id/>
<env/>
<env/>
```

No

```xml
<env/>
<id/>
<env/>
```

oleg-nenashev:
> It's only possible with an exception that the same repeatable elements are placed together.

I see no problem if we set some more restrictive rules in the XSD Schema than the code actually does. `env / id / env` is a use-case which is unlikely used in the wild and, if so, it is quite easy to adjust it.

If the strict schema is enforced in the code at some point, we will likely need a major-digit release. But we firstly need to have a standalone working schema

NextTurn:
My mistake. That exception doesn't help. An XSD 1.0 schema can't be defined unless either one of the conditions in https://github.com/kohsuke/winsw/issues/68#issuecomment-577641310 is met. Otherwise, with the use of `<choice>` and full arrangement we must have 23! = 2e22 nodes in XSD for 23 non-repeatable elements.

PR #369, opened 2020-01-21
Fix configuration samples

### LINKED ISSUE #368: Root element of XML is to be <Service> is not compulsory
In the [xmlConfigFile documentation](https://github.com/kohsuke/winsw/blob/master/doc/xmlConfigFile.md), has mentioned that the root element of the XML configuration file should be **service**.

issue

But the root element of the [sample-allOption.xml](https://github.com/kohsuke/winsw/releases/download/winsw-v2.3.0/sample-allOptions.xml) and [sample-minimal.xml](https://github.com/kohsuke/winsw/releases/download/winsw-v2.3.0/sample-minimal.xml) is **Configuration**.

I think doc should be changed. IIRC the root element name does not really matter.  So its better do edit the documentation.

Thank you!

NextTurn:
Fixes #368

### DIFF
--- a/examples/sample-allOptions.xml
+++ b/examples/sample-allOptions.xml
@@ -24,7 +24,7 @@
  
  You can find more information about configuration options here: https://github.com/kohsuke/winsw/blob/master/doc/xmlConfigFile.md
 -->
-<configuration>
+<service>
   
 <!-- 
 SECTION: Mandatory options
@@ -307,4 +307,4 @@ More info is available here: https://github.com/kohsuke/winsw/blob/master/doc/ex
 </extensions>
 -->
 
-</configuration>
+</service>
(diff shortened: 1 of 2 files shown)


examples/sample-allOptions.xml
```diff
@@ -24,7 +24,7 @@
  
  You can find more information about configuration options here: https://github.com/kohsuke/winsw/blob/master/doc/xmlConfigFile.md
 -->
-<configuration>
+<service>
```
oleg-nenashev: IIRC does not really matter

NextTurn: It's not required by the current implementation. But what about the schema #68?

oleg-nenashev: Fair point

PR #370, opened 2020-01-21
Update xmlConfigFile.md

buddhikac96:
Root element to be <service> is not compulsory.  So that statement removed.


oleg-nenashev:
As per discussion with @NextTurn in https://github.com/kohsuke/winsw/pull/369#discussion_r369651167 , it might make sense to keep the explicit requirement for future versions so that a Schema can be introduced

oleg-nenashev:
Closing it in favor of https://github.com/winsw/winsw/pull/460

PR #372, opened 2020-01-22
Avoid setting current process environment variables

NextTurn:
That test didn't cover the code paths, but it gave you a lot of confidence?


oleg-nenashev:
.NET2 handling looks like a dark magic to me :(
I will need to get to the test instance before merging it

NextTurn:
> .NET2 handling looks like a dark magic to me :(

There is only one .NET 2 implementation so we can assume the layouts and logic as-is.

NextTurn:
Updated to only use the special codes when **running** on .NET 2.0 -3.5.

NextTurn:
Ping @oleg-nenashev

Issue #378, opened 2020-01-26
Fix the WinSW release flow and release 2.4.0


oleg-nenashev:
https://github.com/kohsuke/winsw/commit/6e9d25a7b53a373049936caa34a2fce3edb7afc2 from @NextTurn apparently broke the release job in https://ci.appveyor.com/project/oleg-nenashev/winsw-g2fwp . It uses `appveyor.yml` but sets the arbitrary version. https://github.com/kohsuke/winsw/commit/6e9d25a7b53a373049936caa34a2fce3edb7afc2 sets the version in config-as-code, and hence it overrides the setting in the job.

My next steps:

- [ ] Move the release Pipeline and deployments to configuration-as-code
- [ ] TBD: Switch to the 4-digit versioning scheme (`major.minor.patch.build`)
- [ ] Fix the Pipeline so that it can release the versions
- [ ] Get 2.4.0 released

oleg-nenashev:
Current WinSW Release Job YAML:

```
version: 2.{build}.0
pull_requests:
  do_not_increment_build_number: true
branches:
  only:
  - master
build:
  verbosity: minimal
deploy: off
```

The most of the configs come from appveyor YAML.

Most likely I will change the release flow so that different versioning is applied on tags

NextTurn:
> ```
> version: 2.{build}.0
> pull_requests:
>   do_not_increment_build_number: true
> ```

PR builds use different build numbers.

oleg-nenashev:
Yes, it is also something to be aligned. PR builds and master branch builds are not deployed anywhere at the moment, so the version was not a concern.

I am a bit hesitant about going all-in CD for this repository, but switching to tag-based releases could be useful

oleg-nenashev:
Obviously it may be possible to just switch from AppVeyor to GitHub Actions or Azure DevOps if it becomes a burden. It is not possible right now, because I have no access to repo secrets and hence cannot configure NuGet keys. Fixable if we want to go down this road tho

oleg-nenashev:
GitHub Deployment still fails after #381 . Now it is caused by non-unique File names. Apparently raw names are consumed instead of aliases in the Appveyor;s GitHub deployment:

```
Creating "winsw-v2.4.0" release for repository "kohsuke/winsw" tag "winsw-v2.4.0" commit "9e830f9e883986601a21aef5b8d647b9f4ed8c92"...OK
Uploading "sample-allOptions.xml" to release assets...OK
[...]

PR #379, opened 2020-01-26
Disable version patching in the CI run

NextTurn:
Work item #378 

Please add `patch: true` in release job configuration.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,11 +18,6 @@ before_build:
   # Generates a temporary SNK. Not for real signing
   - cmd: '"C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools\sn.exe" -k winsw_key.snk'
 
-dotnet_csproj:
-  patch: true
-  file: '**\*.csproj'
-  version: $(appveyor_build_version)
-
 build_script:
   - dotnet.exe build src\winsw.sln
   - dotnet.exe publish -f netcoreapp3.1 src\Core\ServiceWrapper\winsw.csproj


appveyor.yml
```diff
@@ -20,7 +19,7 @@ before_build:
   - cmd: '"C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools\sn.exe" -k winsw_key.snk'
 
 dotnet_csproj:
-  patch: true
+  # patch: false
```
oleg-nenashev: It is not going to fix the issue, because we still need to patch the release version. I will try to find alternate way

oleg-nenashev: TL;DR: UI does not override this setting AFAICT

oleg-nenashev:
As a quick fix, I will just duplicate the YAML config

appveyor.yml
NextTurn: > UI does not override this setting AFAICT

Then move the entire `dotnet_csproj:` sections into release configuration?

NextTurn:
PR builds don't need version patching. Is it possible to move the patching task to the release job?

oleg-nenashev:
Yes, will do

oleg-nenashev:
According to my experiments with #380 , copy-pasting files does not work when appveyor.yml is present. I will remove the version field and setup it's injection in the CI job, but I will have to keep the version patch for now

NextTurn:
So you've set up different version patterns for release and non-release jobs and no action is required for the current *appveyor.yml* file?

oleg-nenashev:
> So you've set up different version patterns for release and non-release jobs and no action is required for the current appveyor.yml file?

I have set up a temporary solution, only that. I am not happy about different versioning patterns which may confuse users at some point. I am also not happy about having to store configs in job settings instead of configuration as code. But at least it restores the status quo we had before 2.4.0

casz:
I'd suggest looking into using GitVersion.
See https://github.com/nmklotas/GitLabApiClient the main csproj and the appveyor.yml is of interest.
I can link to other projects with GitVersion in an automated workflow.

GitLabApiClient is fully automated for releasing to nuget.org with versioning when a tag is added.

PR #381, opened 2020-01-26
Fix file extensions for .NET Core artifacts

oleg-nenashev:
CC @NextTurn . It fixes extensions of files which is important for further deployment to GitHub Releases.


oleg-nenashev:
Will merge to finally cut the release this weekend

PR #383, opened 2020-01-27
Produce ready-to-publish artifacts in build phase for AppVeyor

### LINKED ISSUE #382: .NET Core package should not have a double *.zip suffix
See `https://ci.appveyor.com/api/buildjobs/f6p2pfbb6pt9rsw4/artifacts/src%2FCore%2FServiceWrapper%2Fbin%2FRelease%2Fnetcoreapp3.1%2FWinSW.NETCore31.zip.zip` . Looks like the suffix is appended twice during the project build, would be great to get it fixed.

NextTurn:
Fixes #382, https://github.com/kohsuke/winsw/issues/378#issuecomment-578535632

~Removes "deployment name"s since they are not used in GitHub Releases deployment.~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ obj
 /winsw_key.pfx
 /src/.vs/
 /src/Core/WinSWCore/WinSWCore.csproj.DotSettings
+/artifacts/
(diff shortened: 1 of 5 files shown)


oleg-nenashev:
> Removes "deployment name"s since they are not used in GitHub Releases deployment.

Actually they are. The release environment setting uses the deployment names. Should be moved to configuration-as-code somehow, yes:

image

PR #362, opened 2020-01-19
Improve P/Invoke signatures

NextTurn:
Based on #356 

When in doubt, read the docs.

PR #392, opened 2020-01-31
Set description via advapi32

### LINKED ISSUE #208: WinSW should use advapi to write service description
Follow-up to #183. Currently WinSW writes description directly to the registry, which is not guaranteed to work reliably on all Windows versions.

NextTurn:
Fixes #208 

Hold for #362

### DIFF
--- a/src/Core/ServiceWrapper/Main.cs
+++ b/src/Core/ServiceWrapper/Main.cs
@@ -17,7 +17,6 @@ using log4net.Appender;
 using log4net.Config;
 using log4net.Core;
 using log4net.Layout;
-using Microsoft.Win32;
 using winsw.Extensions;
 using winsw.Logging;
 using winsw.Native;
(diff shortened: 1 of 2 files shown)


src/Core/ServiceWrapper/Main.cs
```diff
@@ -642,39 +642,21 @@ public static void Run(string[] _args, ServiceDescriptor? descriptor = null)
                     password,
                     descriptor.ServiceDependencies);
 
-                // update the description
-                /* Somehow this doesn't work, even though it doesn't report an error
-                Win32Service s = svc.Select(d.Id);
-                s.Description = d.Description;
```
oleg-nenashev: I wonder whether it is still an issue on old systems, but I am happy to remove this hack

NextTurn: Do you mean setting description via `Win32_Service`?

oleg-nenashev: Yes. I have no practical way to test on .NET Framework 2 now. But we can go ahead and merge it. It is not a critical functionality, so we can restore this hack for old platforms if needed

NextTurn: The `Win32_Service` approach doesn't work because `Description` is a read-only property.

oleg-nenashev:
@NextTurn just to get on the same page w.r.t pull request labeling for the changelogs, do you see a difference for users here? My labeling approach is that `enhancement` goes to user-facing features while the rest goes to `internal`

NextTurn:
> My labeling approach is that `enhancement` goes to user-facing features while the rest goes to `internal`.

Sounds good. This should be `internal` then.

> Probably I should explicitly add descriptions to labels

Completely agreed.

NextTurn:
Is the linked issue (#208) `internal` or not?

oleg-nenashev:
I would say it is internal. There is a "Currently WinSW writes description directly to the registry, which is not guaranteed to work reliably on all Windows versions." note in the ticket, but so far we had no reports confirming it is a case on modern platforms

PR #397, opened 2020-02-03
Annotate more nullable variables

NextTurn:


oleg-nenashev:
OK, now we got a merge conflict after #374 . Too big one to resolve in WebUI (and my dev env is on another continent)

PR #280, opened 2018-06-20
'If-Modified-Since' support

kvach2avl:
Add support of 'If-Modified-Since' header in request and HTTP/304 in response for resources downloading of the self-updating services

PR #403, opened 2020-02-07
Support 'If-Modified-Since' for downloads

NextTurn:
Supersedes #280

If this is to be an opt-out behavior, here are some proposed XML attribute names:

- `preserveNewest`
- `updateIfNewer`
- `skip(if)Unchanged`/`skip(if)Unmodified`/`skipIfNotMidified`


oleg-nenashev:
Yes, it would still need to be configurable from XML IMHO

PR #407, opened 2020-02-11
Update license

NextTurn:


oleg-nenashev:
well, we can update that

NextTurn:
The simplest way to update the time span is to remove it.

NextTurn:
I wonder why Sun Microsystems is in the list.

oleg-nenashev:
> I wonder why Sun Microsystems is in the list.

Hudson/Jenkins project was started at Sun, and WinSW was originally a part of its codebase. So I believe  the reference comes from there

oleg-nenashev:
what is the purpose of the formatting change BTW?

NextTurn:
> what is the purpose of the formatting change BTW?

It's a magic convention in most MIT licenses that lines do not start with spaces and have a max length of 80 characters...

PR #408, opened 2020-02-12
Update exception handling for downloads

NextTurn:


src/Core/ServiceWrapper/Main.cs
```diff
@@ -211,7 +211,14 @@ protected override void OnStart(string[] args)
                 tasks[i] = download.PerformAsync();
             }
 
-            Task.WhenAll(tasks);
+            try
+            {
+                Task.WaitAll(tasks);
+            }
+            catch
+            {
```
oleg-nenashev: Add some logging here?

NextTurn: The exception caught here is an aggregate of all failed tasks' exceptions. By handling them in another loop, we can link the exceptions with their corresponding downloads. Alternatively we could move the EH logic to the `Download` class as said in the comment.

oleg-nenashev: New version works well for me, thanks!

Issue #210, opened 2017-04-21
Can't start service - WMI Operation failure: AccessDenied


sukinsky:
Hello!
Having problems with actual version on Windows 2012 R2.
Successfully installed service for nginx.exe with default minimal config. Starts with SYSTEM account.

If I start service (under administrator rights) with "net start" I got:
`System error 5 has occurred.
Access is denied.`

If I start service with "WinSW.NET4.exe start" I got:
`2017-04-21 11:35:01,236 INFO  - Starting ServiceWrapper in the CLI mode
2017-04-21 11:35:01,518 FATAL - WMI Operation failure: AccessDenied
WMI.WmiException: AccessDenied
   at WMI.WmiRoot.BaseHandler.CheckError(ManagementBaseObject result)
   at WMI.WmiRoot.InstanceHandler.Invoke(Object proxy, MethodInfo method, Object[] args)
   at winsw.WrapperService.Run(String[] _args, ServiceDescriptor descriptor)
   at winsw.WrapperService.Main(String[] args)
`
Thanks,
sukinsky

oleg-nenashev:
In order to start services from CLI, it is not enough to be a system administrator. Or command shall should be started with "Run as Administrator" privilege.

sukinsky:
Of course I start CMD with "Run as Administrator" privilege. The same error is from Services or Task manager.

oleg-nenashev:
Hm, maybe there are some restrictions for service management in your Windows (e.g. Domain policy). Just to make sure, are you able to manage other services?

sukinsky:
Yes I'm inside enterprise environment with domain and local admin on target computer. I'm able to manage any service installed, I can install own services (may be you know `nssm.exe`).

I used `winsw` on WS2008R2 in same environment without any problems.
Tried add ".NET Framework 3.5 Features", checked that WMI service is running.

sukinsky:
Hi, Oleg!

Old version (2.0.0) it works as a charm. After I tried downgraded version - all versions of winsw began to work. I think there were some local PC issues.

Ishan1987:
HI @sukinsky ,

I had the similar issue that i fixed with below 2 steps:
1) Added user in to admin group
2) goto your downloaded jenkins-slave.exe Application file.
goto properties -> compatibility -.> check run as administrator
Then i retried installing jenkins as service resolved the issue. Make sure to click ok on prompt where service will be run as admin.

ricklefort:
[...]

PR #410, opened 2020-02-15
Remove app manifest file

NextTurn:
This file is not used at all. Created #457 as a replacement.


NextTurn:
Fixes #210

src/Core/ServiceWrapper/app.manifest
```diff
@@ -0,0 +1,10 @@
+﻿<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
+  <assemblyIdentity version="1.0.0.0" name="WinSW" type="win32" />
+  <trustInfo xmlns="urn:schemas-microsoft-com:asm.v2">
+    <security>
+      <requestedPrivileges xmlns="urn:schemas-microsoft-com:asm.v3">
+        <requestedExecutionLevel level="requireAdministrator" uiAccess="false" />
```
oleg-nenashev: uiAccess is `false` for the service mode, but it is an option for the CLI mode

NextTurn: IIUC we'll need a digital signature to enable `uiAccess`.

NextTurn: What does `uiAccess` do?

NextTurn:
NOTE: It's also possible to defer privilege requests until the actual operations, i.e. we don't trigger UAC for `help` or `version`, or if the config file contains invalid data. This will make the codes more complicated but may be more user-friendly.

src/Core/ServiceWrapper/app.manifest
NextTurn: Could you please explain why we need UIAccess? I've got a different direction for this.

oleg-nenashev:
If it is not a big deal to exclude some commands from UAC, +100 for doing so. We may want to add more commands which do not require admin privileges later, so better to keep UX now

PR #418, opened 2020-02-20
Reorganize projects

NextTurn:
#### Before

*ServiceWrapper* build → *WindowsService* publish → *WinSW*

#### After

*WinSW* build → *WinSW* publish → *WinSW*


oleg-nenashev:
I'd guess we need to merge all other PRs before this one

PR #430, opened 2020-03-02
Use generic enum parsing methods in all packages except .NET 2.0

NextTurn:


src/Core/WinSWCore/Util/XmlHelper.cs
```diff
@@ -87,22 +87,33 @@ public static TAttributeType SingleAttribute<TAttributeType>(XmlElement node, st
...
             string rawValue = node.GetAttribute(attributeName);
             string substitutedValue = Environment.ExpandEnvironmentVariables(rawValue);
+#if NET20
             try
             {
                 var value = Enum.Parse(typeof(TAttributeType), substitutedValue, true);
                 return (TAttributeType)value;
             }
-            catch (Exception ex) // Most likely ArgumentException
+            catch (ArgumentException ex)
```
oleg-nenashev: I have no memory of that place

NextTurn: There's only `ArgumentException`:

https://docs.microsoft.com/dotnet/api/system.enum.parse#System_Enum_Parse_System_Type_System_String_System_Boolean_

PR #440, opened 2020-03-15
Remove Runaway Process Killer

NextTurn:
Shifts the responsibility onto the operating system.

This feature is now safely enabled by default.


NextTurn:
@oleg-nenashev There is something wrong with the `ShouldKillTheSpawnedProcess` test. If this is good to go then we don't need to fix the test now.

doc/selfRestartingService.md
```diff
@@ -10,6 +10,6 @@ This hidden command is a flavor of the `restart` operation,
   where winsw creates another winsw process in a separate process group, 
   and restarts the service from there.
 
-This additional indirection is necessary because Windows Service Control Manager (SCM) will kill child processes recursively when it stops a service. 
-SCM doesn't provide the restart operation as an atomic operation either, so winsw implements restart by a sequence of stop and start. 
+This additional indirection is necessary because winsw will kill child processes recursively when stopping.
```
oleg-nenashev, line 13:
```suggestion
This additional indirection is necessary because WinSW will kill child processes recursively when stopping. 
```

src/Test/winswTests/Extensions/RunawayProcessKillerTest.cs
```diff
@@ -50,9 +50,6 @@ public void LoadExtensions()
             // Check the file is correct
             var extension = manager.Extensions["killRunawayProcess"] as RunawayProcessKillerExtension;
             Assert.IsNotNull(extension, "RunawayProcessKillerExtension should be loaded");
```
oleg-nenashev, line 52: Not sure this test has any value anymore

NextTurn:
I suggest moving forward. With a few bug fixes and stability updates WinSW rarely crashes or leaves child processes aside. This will work when a serious crash (CSE/AVE/SOE) or process kill (by a user or SCM) occurs. If the service stops normally (under control) we can turn it off.

- The CREATE_BREAKAWAY_FROM_JOB flag is supported since Windows XP.

oleg-nenashev:
I agree

PR #451, opened 2020-03-24
Update CODEOWNERS to request reviews from all maintainers

oleg-nenashev:
I an use a team or setup individual review requests. What would you prefer @NextTurn ?

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @oleg-nenashev
+* @winsw/winsw-maintainers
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
I will do explicit maintainer mentions then, it is more convenient for external contributors

NextTurn:
Who should the contributors mention (`@`) then if they want to ping on something?

oleg-nenashev:
> Who should the contributors mention (@) then if they want to ping on something?

Individuals. IIRC GitHub still does not allow mentioning public teams if a contributor is not a member of the GitHub organization.

oleg-nenashev:
Fix confirmed in https://github.com/winsw/winsw/pull/448

PR #452, opened 2020-03-25
Add .NET 4.6.1 executables to the NuGet package

oleg-nenashev:
Subj. Since we have an executable now, it is quite straightforward to publish it

### DIFF
--- a/WinSW.nuspec
+++ b/WinSW.nuspec
@@ -27,7 +27,9 @@ More info about the wrapper is available in the projects GitHub repository.
   <files>
     <file src="artifacts\WinSW.NET2.exe" target="lib\net20-full\WinSW.NET2.exe" />
     <file src="artifacts\WinSW.NET4.exe" target="lib\net40-full\WinSW.NET4.exe" />
+    <file src="artifacts\WinSW.NET461.exe" target="lib\net461-full\WinSW.NET461.exe" />
     <file src="examples\sample-allOptions.xml" target="lib\net20-full\WinSW.NET2.xml" />
     <file src="examples\sample-allOptions.xml" target="lib\net40-full\WinSW.NET4.xml" />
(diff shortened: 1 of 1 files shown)


NextTurn:
@oleg-nenashev This doesn't look right that every PR by you now has to wait for my review.

oleg-nenashev:
@NextTurn I am perfectly fine with that as a default behavior. If something is needed to be delivered quickly, I have an escape hatch:

image

We may have an opposite problem though. AFAICT you cannot merge your own PR now if there is no approval from me. I think we need to fix that, I can update the repo settings accordingly.

oleg-nenashev:
Master branch failed with "Build execution time has reached the maximum allowed time for your plan (60 minutes).". Maybe the org account has a more restrictive plan than personal ones, I have not seen 60min timeouts before. Anyway, it is not something common

NextTurn:
It's possibly due to the failed runaway process killer test.

WinSW.nuspec
```diff
@@ -27,7 +27,9 @@ More info about the wrapper is available in the projects GitHub repository.
   <files>
     <file src="artifacts\WinSW.NET2.exe" target="lib\net20-full\WinSW.NET2.exe" />
     <file src="artifacts\WinSW.NET4.exe" target="lib\net40-full\WinSW.NET4.exe" />
+    <file src="artifacts\WinSW.NET461.exe" target="lib\net461-full\WinSW.NET461.exe" />
```
NextTurn: Only .NET 4.0 has the concepts of "Client Profile" and "Full Profile". The problem with `net20-full` and `net40-full` is people may be using hard-coded paths. But we can omit the suffix for `net461`.

oleg-nenashev: Would be nice to update it before the release, will do it. Thanks!

Issue #269, opened 2018-03-02
<depend> is not properly working??


phanigandeed:
In my project, I have a dependency specification like this.
      <executable>java</executable>
      <arguments>-Xrs -Xmx256m -jar "%BASE%\demoservice.jar"</arguments>
      <depend>demoservice2</depend>
      <logmode>rotate</logmode>
    </service>
....

And, I have installed both demoservice, and demoservice2.

But, even if demoservice2 is not running, demoservice is running. As per specification in your documentation, it should not run.

Is my understanding wrong?

oleg-nenashev:
Hello, WinSW just declares dependencies when it installs the service. All management happens on the Windows side

* Which WinSW version do you use?
* What is your operating system?
* Please provide a registry dump for both installed services

NextTurn:
`<depend>` is working properly as of the latest version.

It's possible that the dependency service crashes when running. In that case the dependent services won't automatically stop.

PR #460, opened 2020-03-29
XML schema validation with XSD

### LINKED ISSUE #68: Provide XSD schema for XML configuration files

buddhikac96:
Added XML schema validation for config file. 
I used [this](https://docs.microsoft.com/en-us/dotnet/standard/data/xml/xml-schema-xsd-validation-with-xmlschemaset) approach from .Net docs. 
But validation not working properly. Even though schema is not matched it doesn't throw any error.


oleg-nenashev:
Schema file is missing in this pull request

src/Core/WinSWCore/ServiceDescriptor.cs
```diff
@@ -95,6 +105,30 @@ public ServiceDescriptor(XmlDocument dom)
...
+        private void ValidateXmlSchema()
+        {
+            XmlReaderSettings booksettings = new XmlReaderSettings();
+            booksettings.Schemas.Add(null, BasePath + ".xsd");
+            booksettings.ValidationType = ValidationType.Schema;
+            booksettings.ValidationEventHandler += new ValidationEventHandler(booksSettingsValidationEventHandler);
+
+            XmlReader books = XmlReader.Create(BasePath + ".xml");
+
+            books.Read();
```
NextTurn: `XmlReader.Read` reads only the **next node** but not the entire document.

src/Core/WinSWCore/ServiceDescriptor.cs
```diff
@@ -95,6 +105,30 @@ public ServiceDescriptor(XmlDocument dom)
...
+        private void ValidateXmlSchema()
+        {
+            XmlReaderSettings booksettings = new XmlReaderSettings();
+            booksettings.Schemas.Add(null, BasePath + ".xsd");
+            booksettings.ValidationType = ValidationType.Schema;
+            booksettings.ValidationEventHandler += new ValidationEventHandler(booksSettingsValidationEventHandler);
+
+            XmlReader books = XmlReader.Create(BasePath + ".xml");
+
+            while (books.Read()) { }
```
NextTurn: Now we read the file twice: once here and once in the constructor. Try to reduce the number of reads to one.

buddhikac96: Reading and validation combined

src/Core/WinSWCore/ServiceDescriptor.cs
```diff
@@ -95,6 +100,40 @@ public ServiceDescriptor(XmlDocument dom)
...
+            {
+                dom.Load(reader);
+            }
+        }
+
+        private void XmlValidationEventHandler(object sender, ValidationEventArgs e)
+        {
+            if (e.Severity == XmlSeverityType.Warning)
+            {
+                throw new XmlException("[Warning] XML validation - " + e.Message);
```
NextTurn, line 129: I'm wondering if we need to throw for warnings at all. A warning is typically issued when there is no schema to validate a particular element, e.g. after we add a new setting but do not update the XSD.

buddhikac96: Yes. Its better to move on with just Errors.

src/Core/WinSWCore/ServiceDescriptor.cs
```diff
@@ -95,6 +100,36 @@ public ServiceDescriptor(XmlDocument dom)
             this.dom = dom;
         }
 
+        private void ValidateAndLoadXmlSchema()
+        {
+            XmlReaderSettings settings = new XmlReaderSettings();
+
+            try
+            {
+                settings.Schemas.Add(null, BasePath + ".xsd");
```
NextTurn, line 109: We can't ship the XSD as a physical file. Have a look around [Assembly.GetManifestResourceStream](https://docs.microsoft.com/dotnet/api/system.reflection.assembly.getmanifestresourcestream).

examples/windowsservice.xsd
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0"?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+targetNamespace="https://www.w3schools.com"
+xmlns="https://www.w3schools.com"
+elementFormDefault="qualified">
```
NextTurn, line 5:
```suggestion
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
```

src/Core/WinSWCore/XMLSchema.xsd
```diff
@@ -0,0 +1,18 @@
+﻿<?xml version="1.0"?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+targetNamespace="https://www.w3schools.com"
+xmlns="https://www.w3schools.com"
+elementFormDefault="qualified">
```
NextTurn, line 5:
```suggestion
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
```

src/Core/WinSWCore/ServiceDescriptor.cs
```diff
@@ -65,12 +67,16 @@ public ServiceDescriptor()
...
+                ValidateAndLoadXmlSchema();
             }
             catch (XmlException e)
             {
                 throw new InvalidDataException(e.Message, e);
             }
+            catch (FileNotFoundException e)
+            {
+                throw e;
+            }
```
NextTurn, line 79: Why rethrow?

src/Core/WinSWCore/ServiceDescriptor.cs
```diff
@@ -95,6 +101,46 @@ public ServiceDescriptor(XmlDocument dom)
...
+                    {
+                        settings.Schemas.Add(null, schemaReader);
+                    }
+
+                }
+            }
+            catch(Exception e)
+            {
+                throw e;
+            }
```
NextTurn, line 125: Why rethrow?

nxtn (commented):
Seems like this will work. Can you please add some tests?

buddhikac96:
> Seems like this will work. Can you please add some tests?

Sure.... I will!

src/Core/WinSWCore/ServiceDescriptor.cs
```diff
@@ -95,6 +101,37 @@ public ServiceDescriptor(XmlDocument dom)
...
+                using (XmlReader schemaReader = XmlReader.Create(schemaStream))
+                {
+                    settings.Schemas.Add(null, schemaReader);
+                }
+            }
+
+            settings.ValidationType = ValidationType.Schema;
+            settings.ValidationEventHandler += new ValidationEventHandler(XmlValidationEventHandler);
+            
+            using(XmlReader reader = XmlReader.Create(BasePath + ".xml"))
```
NextTurn, line 121: `BasePath` is `null` in the test because it's not assigned in `ServiceDescriptor.FromXML` but this shouldn't be a blocker. `XmlReader` can be constructed from text as well.

src/Test/winswTests/ServiceDescriptorTests.cs
```diff
@@ -441,5 +444,20 @@ public void DelayedStart_RoundTrip(bool enabled)
...
+            const string seedXml = "<service>" +
+  "<id>myapp</id> " +
+  "<name>MyApp Service (powered by WinSW)</name>" +
+  "<description>This service is a service created from a minimal configuration</description>" +
+"</service>";
+
+            var serviceDescriptor = ServiceDescriptor.FromXML(seedXml);
+
+            var reader = XmlReader.Create(new StringReader(seedXml));
+            serviceDescriptor.ValidateAndLoadXmlSchema(reader);
```
NextTurn, line 460: Now let's focus on the test. The validator's job is to detect invalid XML pieces and throw exceptions for them. In order to verify it works there needs to be a test to ensure that it does throw an exception.

You can write an assertion like `Assert.That(() => DoSomething(), Throws.TypeOf<XmlException>());`.

buddhikac96: Thank you for the description. You are so helping. :)
Here this part is not clear to me. I'm testing here, whether my function which is tested is thrown any **XmlException**. If this throws any Exception then my test will be passed, otherwise test will be failed. So I think I need to feed XML string as the test case, which fails the validation, then exception will be thrown and Test will pass. Am I correct?

NextTurn: Right. You feed an invalid XML config string (valid XML but invalid WinSW config) and expect the validator to throw.

buddhikac96: This is the XSD, I have provided.

<xs:element name="service">
    <xs:complexType>
      <xs:sequence>
        <xs:element name="id" type="xs:string"/>
        <xs:element name="name" type="xs:string"/>
        <xs:element name="description" type="xs:string"/>
        <xs:element name="executable" type="xs:string"/>
      </xs:sequence>
    </xs:complexType>
  </xs:element>

But still validation don't catch error in below XML.

<id>
  <service>myapp</service>
  <name>MyApp Service (powered by WinSW)</name>
  <description>This service is a service created from a minimal configuration</description>
</id>

So how to what kind of a XML will fail in this condition. I think obviously this configuration xml is invalid. But XML validator not catching it.

NextTurn: Hmm, it's definitely odd, right? Just a heads-up, the `XmlReaderSettings` you create won't automatically apply to newly constructed `XmlReader`s.

buddhikac96: Ooh.. I have missed it.  I change method like this.

```
public void ValidateAndLoadXmlSchema(XmlReader reader)
        {
            XmlReaderSettings settings =  new XmlReaderSettings();

            settings.ValidationType = ValidationType.Schema;
            settings.ValidationEventHandler += new ValidationEventHandler(XmlValidationEventHandler);

            Assembly a = Assembly.GetExecutingAssembly();

            using (Stream schemaStream = a.GetManifestResourceStream("winsw.XMLSchema.xsd"))
            {
                using (XmlReader schemaReader = XmlReader.Create(schemaStream))
                {
                    settings.Schemas.Add(null, schemaReader);
                }
            }

            var new_reader = XmlReader.Create(reader, settings);

            dom.Load(new_reader);
        }
```
But still it fails to catch the validation error.

NextTurn: `dom` has been loaded in `ServiceDescriptor.FromXML` so it refuses to load again.

buddhikac96: I add another constructor to ServiceDescriptor and it works. But, since default ctor already there I create it with taking integer as a argument and I pass 0 into that in test and run XML validation method from ServiceDescriptor object created with that constructor. But I don't know whether it is a good practice. Shall I commit that ?

NextTurn:
Let's move the conversions to Gitter.

examples/windowsservice.xsd
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0"?>
```
oleg-nenashev, line 1: No longer needed?

src/Core/WinSWCore/XMLSchema.xsd
```diff
@@ -0,0 +1,140 @@
+﻿<?xml version="1.0"?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
+
+  <xs:element name="service">
```
oleg-nenashev, line 4: It would be great to add some documentation to the fields so that it can be displayed in the XML editor like VS Code

oleg-nenashev (commented):
I think it is mostly ready to be merged. It would be great to add a section about schema validation to https://github.com/winsw/winsw/blob/master/doc/xmlConfigFile.md so that users can discover it. Later we could add a separate section and update the XML examples to reference the schema

src/winsw.sln
```diff
@@ -1,22 +1,23 @@
-﻿Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.31101.0
+﻿
```
NextTurn, line 1: Could you please exclude this file from the PR? I've made some changes in #418.

examples/windowsservice.xsd
buddhikac96: @oleg-nenashev  What sort of details should be added in `xmlConfigFile.md`? I think adding whole schema file will be an overkill. Will just specifying that we are using embedded XSD schema file to validate XML config file and user can follow that schema file be enough ?

src/Core/WinSWCore/XMLSchema.xsd
oleg-nenashev: > @oleg-nenashev What sort of details should be added in xmlConfigFile.md? I think adding whole schema file will be an overkill. Will just specifying that we are using embedded XSD schema file to validate XML config file and user can follow that schema file be enough ?

@buddhikac96 adding a reference to the schema and briefly explaining how to use it?

examples/windowsservice.xsd
oleg-nenashev: See another comment for the response. This comment stands, I am not sure why this additional XSD is needed

src/Core/WinSWCore/XMLSchema.xsd
buddhikac96: Would be great! I will.

.gitignore
```diff
@@ -7,6 +7,14 @@
 *.rsuser
 *.suo
 *.user
+/eval
```
NextTurn, line 10: The current *.gitignore* should have excluded almost everything unnecessary. It's safe to delete files that are not in the list.

src/Core/WinSWCore/XMLSchema.xsd
oleg-nenashev: OK, will wait for it before merging

examples/windowsservice.xsd
oleg-nenashev: ping @buddhikac96

oleg-nenashev (commented):
In the future we should also publish the XSD schema to GitHub Release as a part of the release Pipeline. It will be useful to maintain versioning of the schema

src/Core/WinSWCore/ServiceDescriptor.cs
NextTurn: ping @buddhikac96

nxtn (changes requested):

src/Core/WinSWCore/ServiceDescriptor.cs
```diff
@@ -104,11 +114,39 @@ public ServiceDescriptor(XmlDocument dom)
...
         public static ServiceDescriptor FromXML(string xml)
         {
-            var dom = new XmlDocument();
-            dom.LoadXml(xml);
-            return new ServiceDescriptor(dom);
+            return new ServiceDescriptor(XmlValidation(xml));
         }
 
+
+        public static XmlDocument XmlValidation(string xml)
```
NextTurn, line 121:
```suggestion
        public static XmlDocument XmlValidation(TextReader xml)

```

nxtn (changes requested):

src/winsw.sln
NextTurn: `git checkout master -- src/winsw.sln` will do the work.

doc/xmlConfigFile.md
```diff
@@ -362,3 +362,7 @@ This is useful when the main process is a console, which can respond to Ctrl+C c
 ```xml
 <stopparentprocessfirst>true</stopparentprocessfirst>
 ```
+
+## XML Schema Validation
```
NextTurn, line 366:
```suggestion
## XML schema validation
```

doc/xmlConfigFile.md
```diff
@@ -362,3 +362,7 @@ This is useful when the main process is a console, which can respond to Ctrl+C c
...
+
+We are using embedded [XSD](../src/Core/WinSWCore/XMLSchema.xsd) file to validate XML configurations file schema. You can find sample XSD schema file [here.](../examples/windowsservice.xsd). User can follow this schema file when creating the XML configuration file.
```
NextTurn, line 368:
```suggestion
We are using embedded [XSD](../src/Core/WinSWCore/XMLSchema.xsd) file to validate XML configurations file schema. You can find sample XSD schema file [here.](../examples/windowsservice.xsd). You can follow this schema file when creating the XML configuration file.
```

nxtn (changes requested):

src/Core/WinSWCore/ServiceDescriptor.cs
NextTurn, line 121: > DO give methods names that are verbs or verb phrases.

nxtn (changes requested):

.gitignore
NextTurn: If it wasn't clear enough: could you please revert this file?

oleg-nenashev (changes requested):
The schema may be missing some use-cases, and it may mail on valid configurations. I would suggest to add an escape hatch (e.g. `--ignoreConfigErrors`) or to make the schema validation opt-in. It would allow to not block users if the schema validation fails.

@buddhikac96 do you plan to add documentation to the schema? Or would it be a separate pull request later?

.gitignore
buddhikac96: I prefer that. How could I do that?

NextTurn: `git checkout master -- .gitignore`
`git clean -d -n`

buddhikac96:
> The schema may be missing some use-cases, and it may mail on valid configurations. I would suggest to add an escape hatch (e.g. `--ignoreConfigErrors`) or to make the schema validation opt-in. It would allow to not block users if the schema validation fails.
> 
> @buddhikac96 do you plan to add documentation to the schema? Or would it be a separate pull request later?

If it doesn't matter I will add documentation in separate PR.

.gitignore
buddhikac96: `git clean -d -n`   
This command twice with -n and without -n ?

doc/xmlConfigFile.md
```diff
@@ -362,3 +362,7 @@ This is useful when the main process is a console, which can respond to Ctrl+C c
 ```xml
 <stopparentprocessfirst>true</stopparentprocessfirst>
 ```
+
+## XML schema validation
+
+We are using embedded [XSD](../src/Core/WinSWCore/XMLSchema.xsd) file to validate XML configurations file schema. You can find sample XSD schema file [here.](../examples/windowsservice.xsd). You can follow this schema file when creating the XML configuration file.
```
NextTurn, line 368:
```suggestion
We are using embedded [XSD](../src/Core/WinSWCore/XMLSchema.xsd) file to validate XML configurations file schema. You can find sample XSD schema file [here](../examples/windowsservice.xsd). You can follow this schema file when creating the XML configuration file.
```

.gitignore
NextTurn: Up to you. `-n` is short for `--dry-run` and prevents you from accidentally removing files.

doc/xmlConfigFile.md
```diff
@@ -362,3 +362,7 @@ This is useful when the main process is a console, which can respond to Ctrl+C c
 ```xml
 <stopparentprocessfirst>true</stopparentprocessfirst>
 ```
+
+## XML schema validation
+
+We are using embedded [XSD](../src/Core/WinSWCore/XMLSchema.xsd) file to validate XML configurations file schema. You can find sample XSD schema file [here](../examples/windowsservice.xsd). You can follow this schema file when creating the XML configuration file.
```
NextTurn, line 368:
```suggestion
We are using embedded [XSD](../src/Core/WinSWCore/XMLSchema.xsd) file to validate XML configurations file schema. You can follow this schema file when creating the XML configuration file.
```

There is only one XSD file now.

NextTurn, line 368: > to validate XML configurations file schema

I don't think this is correct. Are we validating the schema or using the schema to validate XML?

src/Test/winswTests/ServiceDescriptorTests.cs
```diff
@@ -441,5 +444,19 @@ public void DelayedStart_RoundTrip(bool enabled)
...
+        {
+            const string seedXml = @"<id>
+  <id>myapp</id>
+  <name>MyApp Service (powered by WinSW)</name>
+  <description>This service is a service created from a minimal configuration</description>
+  <executable>%BASE%\myExecutable.exe</executable>
+</id>
+";
+
+            Assert.That(() => ServiceDescriptor.ValidateXMLSchema(new StringReader(seedXml)), Throws.Exception.TypeOf<XmlException>());
```
NextTurn, line 459:
```suggestion
            Assert.That(() => ServiceDescriptor.ValidateXMLSchema(new StringReader(seedXml)), Throws.TypeOf<XmlException>());

```

doc/xmlConfigFile.md
buddhikac96: Sorry for bad English. Should be change. I'll update.

doc/xmlConfigFile.md
```diff
@@ -362,3 +362,7 @@ This is useful when the main process is a console, which can respond to Ctrl+C c
 ```xml
 <stopparentprocessfirst>true</stopparentprocessfirst>
 ```
+
+## XML schema validation
+
+We are using embedded [XSD schema file](../src/Core/WinSWCore/XMLSchema.xsd) to validate XML configurations file. You can follow this schema file when creating the XML configuration file.
```
NextTurn, line 368:
```suggestion
We are using embedded [XSD schema file](../src/Core/WinSWCore/XMLSchema.xsd) to validate XML configuration files. You can follow this schema file when creating the XML configuration file.
```

src/Test/winswTests/ServiceDescriptorTests.cs
```diff
@@ -441,5 +444,20 @@ public void DelayedStart_RoundTrip(bool enabled)
...
+            const string seedXml = @"<id>
+  <id>myapp</id>
+  <name>MyApp Service (powered by WinSW)</name>
+  <description>This service is a service created from a minimal configuration</description>
+  <executable>%BASE%\myExecutable.exe</executable>
+</id>
+";
+
+            Assert.That(() => ServiceDescriptor.ValidateXMLSchema(new StringReader(seedXml)), Throws.TypeOf<XmlException>());
+
```
NextTurn, line 460:
```suggestion
            Assert.That(() => ServiceDescriptor.FromXML(seedXml), Throws.TypeOf<XmlException>());
```

src/Core/WinSWCore/XMLSchema.xsd
```diff
@@ -0,0 +1,391 @@
...
+              For more information, see the source link. 
+            </xs:documentation>
+          </xs:annotation>
+        </xs:element>
+
+        <xs:element name="depend" type="xs:string" minOccurs="0" maxOccurs="unbounded">
+           <xs:annotation>
+            <xs:documentation xml:lang="en">
+              Specify IDs of other services that this service depends on. 
+              When service X depends on service Y, X can only run if Y is running.
```
NextTurn, line 88: Just a side note, this is misleading. #269

src/Core/WinSWCore/XMLSchema.xsd
```diff
@@ -0,0 +1,391 @@
...
+             This optional element controls the timing in which Windows SCM resets the failure count. 
+             For example, if you specify &lt;resetfailure&gt;1 hour &lt;/resetfailure&gt; and your service continues 
+             to run longer than one hour, then the failure count is reset to zero. This affects the 
+             behavior of the failure actions (see &lt;onfailure&gt; above).
+             In other words, this is the duration in which you consider the service has been running successfully. Defaults to 1 day.
+            </xs:documentation>
+          </xs:annotation>
+        </xs:element>
+
+        <xs:element name="sleeptime" type="xs:string" minOccurs="0" maxOccurs="unbounded">
```
NextTurn, line 214: I suggest removing `sleeptime` and `waithint` from docs as they don't need to be visible to users at all.

nxtn (changes requested):
I want to see more tests before reviewing this.

- Duplicated unique settings.
- Missing required settings.

oleg-nenashev (commented):
Not all fields are documented. @buddhikac96 do you plan to split the PR as we discussed at the meeting on Tuesday?

buddhikac96:
> Not all fields are documented. @buddhikac96 do you plan to split the PR as we discussed at the meeting on Tuesday?

You mean about creating the JSON schema or XSD documentation?

oleg-nenashev:
@buddhikac96 

1. Firstly shipping the XSD schema
2. Then shipping the built-in schema verification

PR #461, opened 2020-03-29
Specifying config file in command line feature added

buddhikac96:
- I create new constructor for ServiceDescriptor which taking filename as a parameter.

- But this constructor is very much same to default constructor rather than get file name from parameter instead API call. It seems like sort of code duplication. With some further work may be able to reduce code duplication.

### DIFF
--- a/src/Core/WinSWCore/Configuration/ServiceAccount.cs
+++ b/src/Core/WinSWCore/Configuration/ServiceAccount.cs
@@ -1,0 +1,10 @@
+﻿namespace winsw.Configuration
+{
+    public class ServiceAccount
+    {
+        public string name;
+        public string domain;
+        public string user;
+        public string allowservicelogon;
+    }
+}
--- a/src/Core/WinSWCore/Configuration/YamlConfiguration.cs
+++ b/src/Core/WinSWCore/Configuration/YamlConfiguration.cs
@@ -1,13 +1,15 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
-
-namespace winsw.Configuration
+﻿namespace winsw.Configuration
 {
     public class YamlConfiguration
     {
-        readonly ServiceAccount? serviceAccount;
-        readonly BasicConfigs? basicConfigs;
+        public string? id { get; set; }
+        public string? name { get; set; }
+        public string? description { get; set; }
+        public string? executable { get; set; }
+        public string? workingdirectory { get; set; }
+
+        public ServiceAccount? serviceaccount { get; set; }
+        public Log? log { get; set; }
     }
 
     public class ServiceAccount
@@ -18,11 +20,13 @@ namespace winsw.Configuration
         public string allowservicelogon { get; set; }
     }
 
-    public class BasicConfigs
+    public class Log
     {
-        public string id { get; set; }
-        public string name { get; set; }
-        public string description { get; set; }
-        public string executable { get; set; }
+        public string sizeThreshold { get; set; }
+        public string keepFiles { get; set; }
+        public string pattern { get; set; }
+        public string autoRollAtTime { get; set; }
+        public string period { get; set; }
+        public string mod { get; set; }
     }
 }
--- a/src/Core/WinSWCore/Configuration/YamlConfiguration.cs
+++ b/src/Core/WinSWCore/Configuration/YamlConfiguration.cs
@@ -1,32 +1,47 @@
-﻿namespace winsw.Configuration
+﻿using System.Collections.Generic;
+
+namespace winsw.Configuration
 {
     public class YamlConfiguration
     {
-        public string? id { get; set; }
-        public string? name { get; set; }
-        public string? description { get; set; }
-        public string? executable { get; set; }
-        public string? workingdirectory { get; set; }
+        public readonly string id;
+        public readonly string? name;
+        public readonly string? description;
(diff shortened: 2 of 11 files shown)


NextTurn:
I think the user should really pass a file **path** so we don't have to recursively search for it...

src/Core/WinSWCore/ServiceDescriptor.cs
```diff
@@ -85,6 +85,43 @@ public ServiceDescriptor()
...
         }
 
+        //This constructor is called when config file specified in command line
+        public ServiceDescriptor(string filepath)
+        {
+            BaseName = "";
+            BasePath = "";
+
+            if (filepath.EndsWith(".xml"))
+                filepath = filepath.Substring(0, filepath.Length - 4);
```
NextTurn: Why is the extension stripped?

buddhikac96: Someone may specify file path with extension and someone may not. To avoid that conflict. Also as in XML schema validation XSD file is also have same name as XML file. So have to make sure load the XML file.

NextTurn: Are you expecting the user to provide the XSD file?

buddhikac96: No. what I did is find the xsd file with the same name of xml file

NextTurn: Firstly, why do you pass to trimmed file path to `File.Exists`?

buddhikac96: Ooh I missed that. That is why File.Exist() method return false. If user specify the complete path with extension then this will be okay. I was thinking user will specify only the file name without extension. Thank you for mention this.

src/Core/ServiceWrapper/Main.cs
```diff
@@ -536,7 +536,14 @@ public static void Run(string[] _args, ServiceDescriptor? descriptor = null)
             bool inCliMode = _args.Length > 0;
 
             // If descriptor is not specified, initialize the new one (and load configs from there)
-            descriptor ??= new ServiceDescriptor();
+            if(_args.Length > 1)
```
NextTurn, line 539: Let me explain how this will behave:

****Without config file specified:****

- The user executes `winsw start`, `inCliMode` evaluates to `true`.
- SCM executes `winsw`, `inCliMode` evaluates to `false`.

****With config file specified:****

- The user executes `winsw start path/to/config`, `inCliMode` evaluates to `true`. `_args[1]` is `start`.
- SCM executes `winsw`, `inCliMode` evaluates to `false`. WinSW doesn't know a config file is specified.

src/Core/WinSWCore/Configuration/ServiceAccount.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace winsw.Configuration
+{
+    public class ServiceAccount
+    {
+        public string name;
+        public string domain;
```
oleg-nenashev, line 6: It is optional according to https://github.com/winsw/winsw/blob/f260d7e5d95a2d50b25a48829b50392cb19c6b65/src/Core/WinSWCore/ServiceDescriptor.cs#L648

src/Core/WinSWCore/Configuration/ServiceAccount.cs
```diff
@@ -0,0 +1,10 @@
...
+        public string user;
+        public string allowservicelogon;
```
oleg-nenashev, line 8: It is optional according to https://github.com/winsw/winsw/blob/f260d7e5d95a2d50b25a48829b50392cb19c6b65/src/Core/WinSWCore/ServiceDescriptor.cs#L652

src/Core/WinSWCore/Configuration/YamlConfiguration.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Collections.Generic;
+
+namespace winsw.Configuration
+{
+    public class YamlConfiguration
+    {
+        public string? id;
```
oleg-nenashev, line 7: Isn't it mandatory for the service operation?

src/Core/WinSWCore/Download.cs
```diff
@@ -29,16 +29,21 @@ public enum AuthType
...
-        public readonly string To;
-        public readonly AuthType Auth;
-        public readonly string? Username;
-        public readonly string? Password;
-        public readonly bool UnsecureAuth;
-        public readonly bool FailOnError;
-        public readonly string? Proxy;
+        public string from;
+        public string to;
+        public AuthType auth;
```
oleg-nenashev, line 34: Should it default to AuthType.none and be optional?

NextTurn: This always defaults to `none`.

src/Core/WinSWCore/Configuration/YamlConfiguration.cs
buddhikac96: Yes. Should be changed

buddhikac96:
This PR is continuing in https://github.com/winsw/winsw/pull/547.  Closing PR.

NextTurn (commented):
There is something wrong with GitHub.

PR #465, opened 2020-03-31
Split commands into functions

NextTurn:
Believe it or not, this accelerates the start-up.


NextTurn:
> Not sure whether it is visible to users.

The function names will appear in a stack trace.

PR #468, opened 2020-03-31
Split improvements and new features

oleg-nenashev:
Maybe is is an overkill, trying to split entries between enhancements and major features which need to be highlighted.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -14,10 +14,10 @@ categories:
     label: removed
   - title: ⚠️ Deprecated
     label: deprecated
-  - title: 🚀 New features and improvements
-    labels:
-      - enhancement
-      - new-feature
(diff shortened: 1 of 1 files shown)


.github/release-drafter.yml
```diff
@@ -14,10 +14,10 @@ categories:
     label: removed
   - title: ⚠️ Deprecated
     label: deprecated
-  - title: 🚀 New features and improvements
-    labels:
-      - enhancement
-      - new-feature
+  - title: ":tada: New features"
```
NextTurn, line 17: #470?

oleg-nenashev: Right. Maybe just sticking with a single label is better

oleg-nenashev:
Will merge it, we can always revert it back later

PR #469, opened 2020-03-31
Enable Settings App for the repository

oleg-nenashev:
Could simplify label and configuration management. It is work-in-progress, not all labels from https://github.com/winsw/winsw/labels have been moved yet. Maybe it is easierto keep labels managed from Web UI though


nxtn (approved):
Interesting.

PR #470, opened 2020-03-31
Issue templates: Fix the Documentation icon rendering

oleg-nenashev:
It does not render in Chrome due to whatever reason, hence I replaced it by unicode emoji


oleg-nenashev:
Hmm... then it is about the page. Anyway, the change should fix it

Issue #435, opened 2020-03-09
Remove strong name signing


NextTurn:
https://github.com/kohsuke/winsw/blob/6918cc2823ab4776707ed5c0cf1e5112f863f0a6/DEVELOPER.md#L9

@oleg-nenashev I don't know about your policies but AFAIK most projects have *.snk* files in their repos.

https://github.com/dotnet/arcade/tree/master/src/Microsoft.DotNet.Arcade.Sdk/tools/snk
https://github.com/microsoft/clrmd/blob/master/eng/Microsoft.Diagnostics.Runtime.snk

---

What's more, we don't need to strong-name at all.

https://github.com/dotnet/runtime/blob/release/5.0-preview3/docs/project/strong-name-signing.md

PR #477, opened 2020-04-03
Set up CI with Azure Pipelines

NextTurn:
The workflows are described on the [Wiki page](https://dev.azure.com/winsw/winsw/_wiki/wikis/winsw.wiki/1/CI-CD-workflows). It's internal right now, but it's unlikely that it needs to be public at all.

How do you use the *.snk* file in CI?

Contributes to #493


oleg-nenashev:
> How do you use the .snk file in CI?

It is pretty much unused except codesigning of the binaries. Well, at least it used to be so in previous build flow versions. https://github.com/winsw/winsw/blob/fef7d7e560066312ba72e4502790828a3fb96346/Directory.Build.props#L5

NextTurn:
> It is pretty much unused except codesigning of the binaries.

Are the binaries signed in the *winsw-release* build?

oleg-nenashev:
> Are the binaries signed in the winsw-release build?

No. Everything should happen in the common build flow. Release flow uses the same appveyor.yml, except the versioning scheme and deployment targets

NextTurn:
Then when do you sign the binaries with the *real* signing key? Automatically or manually?

oleg-nenashev:
There is no signing with a _real_ key at the moment, stub signing has been in place since 2013 or so. My personal signing key cannot be deployed on the release environment, and cutting releases from a local  setup was a no-go option so far. Indeed it should change.

IIRC it was even documented at some point, but I cannot find a trace of it now. Will create a new ticket

NextTurn:
Oh I see. Every release artifact has a different public key! It's worse than none. #435

oleg-nenashev:
Hard to argue. I do not remember a reason why I opted to this approach at that time, but changing it is a right thing to do. Will cut 2.7.0 as is, and then we can tweak the flow for further releases

DEVELOPER.md
```diff
@@ -1,39 +0,0 @@
-# WinSW Developer Information
```
oleg-nenashev, line 1: Could we revert it for now and add a disclaimer that the process is subject to change? Then we can merge it and add patches on the top of it

NextTurn: I don't understand why we put this file in the repo. It's internal for maintainers. External contributors should look into *CONTRIBUTING.md*.

oleg-nenashev: I am fine to rename it to MAINTAINERS.md and move to docs. Will do as a separate PR

oleg-nenashev:
One of the tests failed on Azure DevOps: https://dev.azure.com/winsw/winsw/_build/results?buildId=2&view=logs&j=72b59a5c-f9fa-5437-adfc-3f855724a45f&t=706f6d43-7b36-5d52-2781-608600615bec . I will investigate it later today (after 7PM UTC) or tomorrow. Feel free to tweak the build Pipeline as needed, without waiting for my reviews

NextTurn:
> Feel free to tweak the build Pipeline as needed, without waiting for my reviews

That's a false proposition. I have to wait for your reviews to merge PRs.

oleg-nenashev:
Oh, right. You do not have admin access in the repo. I can get it fixed

NextTurn:
https://github.com/winsw/winsw/blob/bed7b68d0bf339f5d22bcb82e27ee08e33af38b7/src/Test/winswTests/Util/FilesystemTestHelper.cs#L40

```
Wrong line in the parsed Set output file: 
```

That test succeeded in a later run.

NextTurn:
> You do not have admin access in the repo.

It's not a problem of the access.. It's the code owners.

oleg-nenashev:
Admins can override the branch protection rules if needed. Anyway, I relaxed the branch protection for now so that you are not blocked @NextTurn

NextTurn:
> Admins can override the branch protection rules if needed.

I'm not familiar with the user rules on GitHub. That's all for me:

image

PR #478, opened 2020-04-03
Remove strong name signing

### LINKED ISSUE #435: Remove strong name signing

NextTurn:
Closes #435

~- Define `SIGN_ASSEMBLY` environment variable to `true` in release builds.~
~- Move key files to a subdirectory.~

We don't really strong-name assemblies at the moment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -239,8 +239,7 @@ orleans.codegen.cs
 
 # Including strong name files can present a security risk
 # (https://github.com/github/gitignore/pull/2483#issue-259490424)
-*.snk
-*.pfx
+#*.snk
 
 # Since there are multiple workflows, uncomment next line to ignore bower_components
 # (https://github.com/github/gitignore/pull/1529#issuecomment-104372622)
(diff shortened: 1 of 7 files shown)


NextTurn:
Merging to unblock the `[InternalsVisibleTo]` attribute. We can restore it anytime once we have real strong-name signing.

PR #479, opened 2020-04-03
Allowing for extra whitespace padding in the <arguments> xml entry

scottselberg:
I want to use WinSW to setup 'swarm-client' to connect into Jenkins.  The swarm-client has a lot of configuration arguments, so it would be really helpful to pad the xml with newlines and whitespace for readability.  The code changes I'm proposing would collapse the whitespace, newlines and line returns down to a single whitespace so the process call works correctly and things report nicely in the logs.

Example xml: 
```
<service>
    <id>jenkinsswarmclient</id>
    <name>Jenkins Swarm Client (powered by WinSW)</name>
    <description>This service starts up the java swarm-client
                 to connect this computer as a jenkins slave node.</description>
    <executable>java</executable>
    <arguments>-jar "%BASE%\swarm-client.jar"
               -master https://jenkins.it.keysight.com  
               -username my-user 
               -password <redacted> 
               -deleteExistingClients 
               -fsroot C:\jenkins -labels "vsbuild windows" 
               -disableClientsUniqueId 
               -name Win-29 
               -mode exclusive 
               -executors 4 
               -description "swarm-client" 
               -retryBackOffStrategy exponential 
               -maxRetryInterval 5
    </arguments>
</service>
```

### DIFF
--- a/src/Core/ServiceWrapper/Main.cs
+++ b/src/Core/ServiceWrapper/Main.cs
@@ -8,6 +8,7 @@ using System.Runtime.InteropServices;
 using System.Security.AccessControl;
 using System.ServiceProcess;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading;
 #if VNEXT
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


NextTurn:
I'm in favor of this idea, but collapsing all white spaces may lead to unexpected results for spaces in arguments. Could this only collapses line breaks and leading/trailing white spaces?

scottselberg:
Hi,  thanks for considering this.

I added the collapsing of the white space because when I created the config for swarm-client, I wanted to line up all the arguments - which means if I did not collapse them I would have a ton of spaces between each arguments.  I believe that would be perfectly functional, but be rather ugly in the logs.  

I don't think I've ever seen a command where it wanted to see two or more spaces.  So that's where it seemed like a better option was just to collapse all white space to one.  I also used the regex because I was first getting rid of new lines and returns, then collapsing spaces, then I realized my editor was tossing in tabs.  So it just seemed a lot safer and more supportable to just collapse all white space using the regex.

So, that's my argument to leave it as is.  If you still have concerns about collapsing the white space, just drop another note and I'll update the code to remove new lines, tabs and line returns which is what I think is required to make the multi-line argument functional.

PR #483, opened 2020-04-06
Feature/eliminate newlines

scottselberg:
Hi,

I created a branch that would just remove newlines, line returns and tabs from the program arguments.  In reviewing the log files, I could see no difference between this code and the one that collapsed whitespace.  I intentionally inserted multiple whitespaces in both quoted and unquoted sections of the arguments, and only saw a single whitespace in the event log.  When testing, I had inserted a event log entry stating that newlines were being removed just to make sure the new code was actually running, and I was seeing that log message.  So, I feel pretty confident that somewhere else in the chain, and it's likely not code that either you or I own, there is a call to collapse whitespace.  I could even be the log viewer that doing the collapsing.

As I'm not seeing any observable differences between the code with the call to collapse whitespace and the one that does't, I'll be very happy to see either in a new release.  So, feel free to accept whichever of my pull requests you feel is the better option and deny the other.

Replaces #479

### DIFF
--- a/src/Core/ServiceWrapper/Main.cs
+++ b/src/Core/ServiceWrapper/Main.cs
@@ -8,6 +8,7 @@ using System.Runtime.InteropServices;
 using System.Security.AccessControl;
 using System.ServiceProcess;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading;
 #if VNEXT
 using System.Threading.Tasks;
@@ -276,12 +277,11 @@ namespace winsw
                 startarguments += " " + _descriptor.Arguments;
             }
 
-            // Collapsing newlines, line returns, tabs and multiple spaces into a single
-            // space.  This allows users to pad the arguments in the xml for readability
-            // without breaking the application launching and keeps the log entries 
-            // compact.
-            startarguments = startarguments.Replace("\t", " ").Replace("\n", " ").Replace("\r", " ");
-            startarguments = string.Join(" ", startarguments.Split(new char[] {' '}, StringSplitOptions.RemoveEmptyEntries));
+            // Collapsing whitespace such as newlines, line returns, tabs and multiple 
+            // spaces into a single space.  This allows users to pad the arguments in 
+            // the xml for readability without breaking the application launching and 
+            // keeps the log entries compact.
+            startarguments = Regex.Replace(startarguments, @"\s+", " ");
 
             LogEvent("Starting " + _descriptor.Executable + ' ' + startarguments);
             Log.Info("Starting " + _descriptor.Executable + ' ' + startarguments);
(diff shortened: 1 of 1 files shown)


src/Core/ServiceWrapper/Main.cs
```diff
@@ -276,6 +277,11 @@ protected override void OnStart(string[] args)
                 startarguments += " " + _descriptor.Arguments;
             }
 
+            // Converting newlines, line returns, tabs into a single 
+            // space. This allows users to provide multi-line arguments
+            // in the xml for readability.
+            startarguments = Regex.Replace(startarguments, @"[\n\r\t]+", " ");
```
oleg-nenashev, line 283:
```suggestion
            startarguments = Regex.Replace(startarguments, @"\s+[\n\r\t]+\s+", " ");

```

oleg-nenashev: I'd guess it finalizes the proposal from @NextTurn in https://github.com/winsw/winsw/pull/479#issuecomment-608967443

NextTurn: Indeed this is what I thought in https://github.com/winsw/winsw/pull/479#issuecomment-608967443. **This is acceptable.**

But now I'm thinking of a more conservative approach, which is to collapse only white spaces that are treated as delimiters by Windows.

Don't feel that you have to address it. This is what I planned for v3. In v2 you can use `<argument>`/`<startargument>`/`<stopargument>` without `s`.

(I know no one wants to put line breaks in command line arguments. But it's completed valid, unless we request users not to split one argument into lines.)

scottselberg: @oleg-nenashev - the replacement pattern `@"\s+[\n\r\t]+\s+"` will only work if you have padding before and after the newline|linereturn|tab character.  So if you just have your argument followed by a newline, which I think will be a common case, that pattern will not match and remove the newline.

NextTurn: Should be @"\s*[\n\r\t]+\s*".

scottselberg: Ah, that makes a lot more sense.

scottselberg: Pull request updated.

src/Core/ServiceWrapper/Main.cs
```diff
@@ -276,6 +277,11 @@ protected override void OnStart(string[] args)
                 startarguments += " " + _descriptor.Arguments;
             }
 
+            // Converting newlines, line returns, tabs into a single 
+            // space. This allows users to provide multi-line arguments
+            // in the xml for readability.
+            startarguments = Regex.Replace(startarguments, @"\s*[\n\r\t]+\s*", " ");
```
NextTurn, line 283:
```suggestion
            startarguments = Regex.Replace(startarguments, @"\s*[\n\r]+\s*", " ");
```

Tabs aren't special.

scottselberg: Updated the pull request.

PR #485, opened 2020-04-07
Redirect event logging

NextTurn:
Removes duplicated warning/error entries.


oleg-nenashev:
I need to investigate that a bit. Call duplication was deliberate when I was moving logging to log4net, but I do not recall what was the reason. Not sure this is still actual

NextTurn:
> Call duplication was deliberate when I was moving logging to log4net, but I do not recall what was the reason.

I see. It must be due to the log level.

https://github.com/winsw/winsw/blob/31af6fe11c75f6f3d952446023abb484b717a792/src/Core/ServiceWrapper/Main.cs#L885

PR #491, opened 2020-04-08
Enable system default TLS versions when targeting .NET Framework 4.6.1

NextTurn:
See jenkinsci/windows-slave-installer-module#25 for background.

This should enable TLS 1.3 on newer Windows versions.


NextTurn:
As per the docs an experimental implementation of TLS 1.3 is included in Windows 10, version 1909. It's disabled by default.

[What's new in Windows 10, version 1909](https://docs.microsoft.com/windows/whats-new/whats-new-windows-10-version-1909#transport-layer-security-tls)

PR #492, opened 2020-04-08
Move maintainer documentation to a directory under /doc/maintainers

oleg-nenashev:
subj. Follow-up to the discussion in #477


oleg-nenashev:
We will merge it if we do not migrate to Azure DevOps soon enough. Otherwise I will rewrite the pull request before merging

oleg-nenashev:
This change is no longer needed. We move flows to Azure DevOps, and a new documentation is needed. I will try to write one tonigs

PR #502, opened 2020-04-11
Always set process priority if specified

### LINKED ISSUE #169: Process priority parameter - strange logic
Service descriptor returns `null` priority if the setting is not specified
```
           var priority = _descriptor.Priority;
            if (priority != ProcessPriorityClass.Normal)
                processToStart.PriorityClass = priority;
```

So the actual behavior is:
- If the property is not specified, null is set. According to MSDN, it defaults to `Normal`
- If the property is `Normal`, PriorityClass is not assigned even if the initial priority was different
- Otherwise, the value is assigned

NextTurn:
Fixes #169 

According to the docs:

> By default, the priority class of a process is NORMAL_PRIORITY_CLASS. Use the CreateProcess function to specify the priority class of a child process when you create it. If the calling process is IDLE_PRIORITY_CLASS or BELOW_NORMAL_PRIORITY_CLASS, the new process will inherit this class.

### DIFF
--- a/src/Core/WinSWCore/Util/ProcessHelper.cs
+++ b/src/Core/WinSWCore/Util/ProcessHelper.cs
@@ -183,7 +183,7 @@ namespace winsw.Util
             processToStart.Start();
             Logger.Info("Started process " + processToStart.Id);
 
-            if (priority != null && priority.Value != ProcessPriorityClass.Normal)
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Better to just get it fixed

PR #510, opened 2020-04-14
Rework Shared Directory Mapper

### LINKED ISSUE #307: can not find network Hard Disk
I have Mounted one network Hard Disk for windows server2012,and when I use winsw to regist service for my spring boot app,it can not find my network Hard Disk.but if I start my app by "java -jar xxx.jar" at  command line, the network Hard Disk can be found.

NextTurn:
Closes #307 

*net1.exe* (which implements `net share`) has these APIs in its import table so I'm pretty sure this is how it works.


NextTurn:
*net1.exe* (which implements `net share`) has these APIs in its import table so I'm pretty sure this is how it works.

NextTurn:
Ping @oleg-nenashev

PR #513, opened 2020-04-16
Reorganize interop codes

NextTurn:
**TL;DR** No functional changes.

We're using more Windows APIs. I think splitting interop codes by function helps to understand.

- Splits `Kernel32` into `ProcessApis` and `ConsoleApis`.
- Splits `Advapi32` into `ServiceApis` and `SecurityApis`.
- Normalizes exception messages to "Failed to do something. {[FormatMessage](https://docs.microsoft.com/windows/win32/api/winbase/nf-winbase-formatmessage)}".
- Changes `Failed to detach from console.` to `Failed to attach to console.` which generates more noise for now but will be eliminated in a following update.


NextTurn:
Removes the problematic version comparison. This counts a bug fix.

https://github.com/winsw/winsw/blob/ea6cb8062aa0c2479cbbaec107bdb6050b110f1a/src/Core/WinSWCore/Native/Advapi32.cs#L121-L125

Issue #493, opened 2020-04-08
Move CI/CD Pipelines to Azure DevOps


oleg-nenashev:
As discussed with @NextTurn , we would like to move the CI setup to Azure DevOps which is more suited for .NET CI/CD at the moment. It would help us to resolve issues with the versioning of the CI builds and to offer a CD version of WinSW later

oleg-nenashev:
@NextTurn Just FYI, I configured the build Pipeline on Azure DevOps: https://dev.azure.com/winsw/winsw/_build?definitionId=1&_a=summary . I added variables there and ran the bit as `2.8.0.1` (which is a wrong version, but fine for testing)

NextTurn:
After #494 the variables will be fetched from https://dev.azure.com/winsw/winsw/_library?itemType=VariableGroups&view=VariableGroupView&variableGroupId=1&path=Version

NextTurn:
The *winsw-release* build of AppVeyor is marking every PR as failed.

oleg-nenashev:
> The winsw-release build of AppVeyor is marking every PR as failed.

It should be fixed now

oleg-nenashev:
Just to follow-up on the changelogs and release flow discussion, I think we could generate fancy changelogs in Markdown on Azure DevOps. Today I hacked a PoC for Jenkins Core Changelog generator which does the job in a single ruby file.

Similar code can be added to WinSW release Pipelines while we wait for a full CLI tool in https://github.com/release-drafter/release-drafter/issues/148

oleg-nenashev:
I configured the NuGet flow and slightly modified the GitHub Releases one.
Currently produced artifacts look to be good, version metadata is also fine.

![image](https://user-images.githubusercontent.com/3000480/79673578-51e4ee00-81db-11ea-8b3e-ec13a965117a.png)

![image](https://user-images.githubusercontent.com/3000480/79673591-63c69100-81db-11ea-84e6-471cbacdb66d.png)

@NextTurn if you are fine with that, I suggest to ship the 2.8.0 release. GitHub Release Draft with changelogs is ready: https://github.com/winsw/winsw/releases/tag/untagged-d2e2868145da865a074a . I will deploy NuGet after GitHub is published

NextTurn:
> I think we could generate fancy changelogs in Markdown on Azure DevOps.

Sounds cool. 

You can ship 2.8.0 at any time, or maybe integrate #457 first.

[...]

PR #525, opened 2020-04-26
Stop building the project on Appveyor, remove the build definition

oleg-nenashev:
Build and release flows are now hosted on Azure DevOps. We do not longer need the AppVeyor definition

Contributes to #493


oleg-nenashev:
All old credentials have been also removed

PR #535, opened 2020-05-17
Decouple CLI and service

### LINKED ISSUE #83: Move CLI parsing and Service host code to different classes
Hard to tell what is going on in Main.cs since it both does service hosting and CLI parsing.

Also remove as much functionality as possible from Run() to make it easier to test from UnitTests.
This should be in scope for 2.0

NextTurn:
Fixes #83 

FYI


oleg-nenashev:
+100, as long as it works

PR #547, opened 2020-06-16
Command line test

buddhikac96:


src/Core/ServiceWrapper/CLI/InstallOption.cs
```diff
@@ -0,0 +1,17 @@
+﻿using CommandLine;
+
+namespace winsw.CLI
+{
+    [Verb("install", HelpText = "Install Windows Service Wrapper")]
+    public class InstallOption : CliOption
+    {
+        [Option("configFile", Required = false, HelpText = "Specify the Configuration file")]
+        public string configFile { get; set; }
```
oleg-nenashev, line 9: Should it be generalized in a base class? Many commands will need this parameter

buddhikac96: Updated

src/Core/ServiceWrapper/Program.cs
```diff
@@ -124,67 +142,55 @@ public static void Run(string[] argsArray, ServiceDescriptor? descriptor = null)
                 elevated = IsProcessElevated();
             }
 
-            switch (args[0].ToLower())
+            switch (obj)
```
NextTurn, line 145: 💭 Is there no better way than using `switch`? Maybe polymorphism.

buddhikac96: Would be great. But isn't it reduce the readability?

buddhikac96: Something that I can guess is method overloading. But I think it will reduce the code readability?

NextTurn: Why does method overloading reduce the readability? Why was OOP invented?

NextTurn: Method overriding.

buddhikac96: Take local functions of commands (`install`, `start` . . . ) into particular CLI utility class with a common name like `RunCommand` and call that method inside the `Run` method?

NextTurn: https://docs.microsoft.com/dotnet/csharp/programming-guide/classes-and-structs/polymorphism

src/Core/ServiceWrapper/CLI/CliOption.cs
```diff
@@ -0,0 +1,65 @@
...
+using System;
+using System.ComponentModel;
+using System.Diagnostics;
+using System.IO;
+using winsw.Native;
+using WMI;
+
+namespace winsw.CLI
+{
+    public abstract class CliOption
```
oleg-nenashev, line 11: This class contains not only an option, but also the execution implementation. I would recommend calling it `CLICommand`, and also renaming other classes in a similar way

src/Core/ServiceWrapper/CLI/StopWaitOption.cs
```diff
@@ -0,0 +1,41 @@
+﻿using CommandLine;
+using System.Threading;
+using WMI;
+
+namespace winsw.CLI
+{
+    [Verb("stopwait", HelpText = "stop the service and wait until it's actually stopped")]
+    public class StopWaitOption : CliOption
```
oleg-nenashev, line 8: There is a lot of duplication with the `StopOption`  Maybe one class could inherit another and slightly extend the method

src/Core/ServiceWrapper/CLI/TestWaitOption.cs
```diff
@@ -0,0 +1,27 @@
+﻿using CommandLine;
+using System;
+using WMI;
+
+namespace winsw.CLI
+{
+    [Verb("testwait", HelpText = "starts the service and waits until a key is pressed then stops the service")]
+    public class TestWaitOption : CliOption
+    {
+        public override void Run(ServiceDescriptor descriptor, Win32Services svcs, Win32Service? svc)
```
oleg-nenashev, line 10: Same, a lot of duplication

src/Test/winswTests/winswTests.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net40;net461;netcoreapp3.1</TargetFrameworks>
```
oleg-nenashev, line 4: I am still waiting to see a prototype of how it resolves the .NET 4.0 support requirements in the binaries

src/Core/ServiceWrapper/CLI/TestWaitOption.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            if (!Program.elevated)
+            {
+                Elevate();
+                return;
+            }
+
+            var arguments = Parser.Default.FormatCommandLine(Program.cliOption).Split(' ');
```
NextTurn, line 18: This doesn't work if the path contains spaces.

src/Core/ServiceWrapper/CLI/TestOption.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        public override void Run(ServiceDescriptor descriptor, Win32Services svcs, Win32Service? svc)
+        {
+            if (!Program.elevated)
+            {
+                Elevate();
+                return;
+            }
+
+            var arguments = Parser.Default.FormatCommandLine(Program.cliOption).Split(' ');
```
NextTurn, line 18: This doesn't work if the path contains spaces.

src/Core/ServiceWrapper/winsw.csproj
```diff
@@ -20,7 +20,14 @@
     <PackageReference Include="log4net" Version="2.0.8" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="CommandLineParser" Version="2.8.0" />
+  </ItemGroup>
+
   <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
+    <PackageReference Include="CommandLineParser">
```
NextTurn, line 28: Duplicate package reference.

src/Core/ServiceWrapper/winsw.csproj
```diff
@@ -41,6 +48,12 @@
     <ProjectReference Include="..\..\Plugins\SharedDirectoryMapper\SharedDirectoryMapper.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
+    <PackageReference Include="CommandLineParser">
```
NextTurn, line 52: ditto

src/Core/ServiceWrapper/winsw.csproj
```diff
@@ -1,8 +1,8 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
NextTurn, line 1: Please revert BOM changes.

src/Core/ServiceWrapper/CLI/CliOption.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        [Option("configFile", HelpText = "Configurations File")]
+        public string ConfigFile { get; set; }
+
+        [Option("elevated", HelpText = "Elevated Command Prompt", Default = false)]
+        public bool Elevated { get; set; }
+
+        [Option("redirect", HelpText = "Redirect Logs")]
```
NextTurn, line 19: :warning: Breaking change: `CommandLineParser` is not going to support forward slashes.

src/Core/ServiceWrapper/CLI/InstallOption.cs
```diff
@@ -0,0 +1,135 @@
...
+using System.Text;
+using winsw.Native;
+using WMI;
+
+namespace winsw.CLI
+{
+    [Verb("install", HelpText = "install the service to Windows Service Controller")]
+    public class InstallOption : CliOption
+    {
+        [Option('p', "profile", Required = false, HelpText = "Service Account Profile")]
```
NextTurn, line 14: :warning: Breaking change: `CommandLineParser` is not going to support forward slashes.

src/Core/ServiceWrapper/winsw.csproj
buddhikac96: what are the BOM changes? I don't know that :(

src/Core/ServiceWrapper/CLI/StopWaitOption.cs
buddhikac96: When overriding the method doesn't I have to implement whole method in again subclass? Is there any pattern on something I can follow to do that?

src/Core/ServiceWrapper/CLI/TestOption.cs
buddhikac96: As I know internally those arguments are not used. I doubt that which kind of argument array will be fit to this. At the moment this is the argument structure.
`winsw install --configFile <filepath> --elevated true -s true`

How should the items be separated here. I found in the commandline library. There is no way to unprase the object back into arguments array. So I think I have to manually convert it?

src/Core/ServiceWrapper/winsw.csproj
buddhikac96: Updated this. Review will help!

src/Core/ServiceWrapper/CLI/TestOption.cs
NextTurn: > internally those arguments are not used.

OK. Then passing an empty array would be preferred.

src/Core/ServiceWrapper/winsw.csproj
NextTurn: Simply revert this line.

src/Core/ServiceWrapper/CLI/CliOption.cs
NextTurn: This argument has been deprecated. We don't need a new one for it.

src/Test/winswTests/winswTests.csproj
buddhikac96: Update in https://github.com/winsw/winsw/pull/547/commits/cfacc38720922d1068f4334cba1c81936c069d62. Review will help to proceed!

NextTurn: @buddhikac96 We still support net40 in v2.

src/Core/ServiceWrapper/CLI/TestOption.cs
buddhikac96: Make sense! I will do so.

src/Core/ServiceWrapper/CLI/CliOption.cs
buddhikac96: I think this deprecating mean we just don't allowing users to elevate manually and still we elevate internally before execute the command? So is this update to Program class works?
```c#
            if (Environment.OSVersion.Version.Major == 5)
            {
                // Windows XP
                elevated = true;
            }
            else
            {
                elevated = IsProcessElevated();
            }
```

src/Core/ServiceWrapper/Program.cs
```diff
@@ -23,18 +23,38 @@
...
+                {
+                    Console.WriteLine("Invalid command [" + args[0] + "]");
+                    PrintAvailableCommands();
+                    return -1;
+                }
+
+                var types = LoadVerbs();
+
+                Parser.Default.ParseArguments(args, types)
+                        .WithParsed(RunParsed)
```
NextTurn, line 55: I just took a look at the docs. Why don't you use the generic version of `WithParsed` in the first place?

oleg-nenashev:
IIUC it would be better to re-target this pull request towards v3

NextTurn:
I thought this was for v2?

oleg-nenashev:
> I thought this was for v2?

Well, we still have an issue with .NET Framework 2.0 support and incompatible CLI changes here. Moving it to v3 would resolve these concerns alltogether

Legion2:
Can winsw be installed globally in PATH on the system and used as a CLI to install multiple services?

buddhikac96:
> Can winsw be installed globally in PATH on the system and used as a CLI to install multiple services?

Yes. But you need to have a separate configuration file for each service that you going to install and after this New CLI feature released then you can specify the configuration file explicitly in the command line interface. Until that, your configuration file should be located in the same directory where WinSW executable is located or in an ancestor directory. That means technically until this gets released you have to manually manage configuration files by renaming or save in a separate folder.

PR #553, opened 2020-06-23
Improve signal-related logging

NextTurn:


NextTurn:
Tuples are not supported on older frameworks... Let me use `KeyValuePair` instead. :(

PR #558, opened 2020-06-26
Add a Dependabot configuration

oleg-nenashev:
Adds Dependabot for NuGet dependencies. Let's see whether it will pick up the current format

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  # Enable version updates for npm
+  - package-ecosystem: "nuget"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "nuget"
+    directory: "/"
```
NextTurn, line 4: 💭 This may need to be a full path.

oleg-nenashev: Might be. I have never used it for .NET before.
Let's try and see

PR #569, opened 2020-07-06
Remove unnecessary project items

NextTurn:


oleg-nenashev:
Why are they unnecessary ?

NextTurn:
Package references flow to next projects so we don't need to define them multiple times.

PR #570, opened 2020-07-08
Standardize coding styles

NextTurn:


NextTurn:
The analyzer will tell people what to do.

Issue #544, opened 2020-06-05
Update codebase to use IWinSWConfiguration instead of ServiceDescriptor internally


oleg-nenashev:
There is a new configuration class which generalizes configuration API: https://github.com/winsw/winsw/blob/master/src/Core/WinSWCore/Configuration/IWinSWConfiguration.cs . Unfortunately the code itself does not use it, `ServiceDescriptor` is passed directly to the internal logic.

I suggest refactoring the code so all of it uses `IWinSWConfiguration` internally. It will help to add YAML support, because we won't need to worry about anything except API. It should help #543 

Implementation notes:

* Extension API uses ServiceDescriptor. It might make sense to keep binary compatibility and to introduce new methods there. https://github.com/winsw/winsw/blob/master/src/Core/WinSWCore/Extensions/AbstractWinSWExtension.cs

PR #543, opened 2020-06-02
YAML Configurations Support (GSoC - 2020)

buddhikac96:
PR contains, 
 - YAML Parsing and Deserializing - https://github.com/winsw/winsw/issues/436
 - IWinSWConfiguration Support - https://github.com/winsw/winsw/issues/544

`YAMLConfiguration` class Is the Utility class for YAML Deserialization.
`ServiceDescriptorYaml` class responsible of find the YAML file and deserialize into object graph.

Previously WinSW supported for `ServiceDescriptor` and in this PR WinSW updated to `IWinSWConfiguration` Support. Now can use any configuration which implements `IWinSWConfiguration`.

PR #573, opened 2020-07-12
Update codebase to use IWinSWConfiguration instead of ServiceDescriptor internally

buddhikac96:
Related Issue : https://github.com/winsw/winsw/issues/544

In current implementation of WinSW it is not possible to use any other configuration type since Program is strictly requiring `ServiceDescriptor` type.  So when YAML Configuration implemented it is not possible to use both XML and YAML configuration at the same time. So in this PR Update Program to requiring `IWinSWConfiguration` type which is the base type of both XML and YAML configurations.

Work Done Yet

- [x] - IWinSWConfiguration Support
- [ ] - Update Unit Test

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "nuget"
+    directory: "/"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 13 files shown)


buddhikac96:
Implement in https://github.com/winsw/winsw/pull/543. Closing the PR.

PR #596, opened 2020-07-22
Update Yaml Configuration support

buddhikac96:
**Updates**
- Add sample-allOption.yml (There are few differences in config names compared to XML configuration to keep UX)
- Update time span values to parsing from strings. Now user can specify TimeSpan values in `5 sec` manner.
- Implement unimplemented properties in YamlConfiguration.cs
- Remove unnecessary properties from ServiceDescriptorYaml.cs

**Possible Updates**
- Document the configurations is sample-allOption.yml file

 **On-going updates**
- WinSW YAML Support Documentation.

### DIFF
--- a/doc/YamlConfigFile.md
+++ b/doc/YamlConfigFile.md
@@ -16,14 +16,13 @@ id: jenkins
 name: Jenkins
 description: This service runs Jenkins continuous integration system.
 env:
-    -
-        name: JENKINS_HOME
-        value: '%BASE%'
+    - name: JENKINS_HOME
+      value: '%BASE%'
 executable: java
 arguments: >
-    -Xrs 
-    -Xmx256m 
-    -jar "%BASE%\jenkins.war" 
+    -Xrs
+    -Xmx256m
+    -jar "%BASE%\jenkins.war"
     --httpPort=8080
 log:
     mode: roll
@@ -104,7 +103,7 @@ See the [Logging and error reporting](loggingAndErrorReporting.md) page for more
 `arguments` element specifies the arguments to be passed to the executable. User can specify all the commands as a single line.
 
 ```yaml
-args: arg1 arg2 arg3
+arguments: arg1 arg2 arg3
 ```
 
 Also user can specify the arguemtns in more structured way with YAML multline strings.
--- a/doc/YamlConfigFile.md
+++ b/doc/YamlConfigFile.md
@@ -1,0 +1,399 @@
+# YAML configuration file
+
+This page describes the configuration file, which controls the behavior of the Windows service.
+
+You can find configuration file samples in the [examples](../examples) directory of the source code repository.
+Actual samples are being also published as a part of releases on GitHub and NuGet.
+
+## File structure
+
+YAML Configuration file shuold be in following format
+
+Example:
+
+```yaml
+id: jenkins
+name: Jenkins
+description: This service runs Jenkins continuous integration system.
+env:
+    -
+        name: JENKINS_HOME
+        value: '%BASE%'
+executable: java
+arguments: >
+    -Xrs 
+    -Xmx256m 
+    -jar "%BASE%\jenkins.war" 
+    --httpPort=8080
+log:
+    mode: roll
+```
+
+## Environment variable expansion
+
+Configuration YAML files can include environment variable expansions of the form `%Name%`. 
+Such occurrences, if found, will be automatically replaced by the actual values of the variables. 
+If an undefined environment variable is referenced, no substitution occurs.
+
+Also, the service wrapper sets the environment variable `BASE` by itself, which points to a directory that contains the renamed *WinSW.exe*. 
+This is useful to refer to other files in the same directory. 
+Since this is an environment variable by itself, this value can be also accessed from the child process launched from the service wrapper.
+
+## Configuration entries
+
+### id
+
+Specifies the ID that Windows uses internally to identify the service.
+This has to be unique among all the services installed in a system,
+  and it should consist entirely out of alpha-numeric characters.
+
+### name
+
+Short display name of the service, which can contain spaces and other characters.
+This shouldn't be too long, like `id`, and this also needs to be unique among all the services in a given system.
+
+### description
+
+Long human-readable description of the service.
+This gets displayed in Windows service manager when the service is selected.
+
+### executable
+
+This element specifies the executable to be launched.
+It can be either absolute path, or you can just specify the executable name and let it be searched from `PATH` (although note that the services often run in a different user account and therefore it might have different `PATH` than your shell does.)
+
+### startmode
+
+This element specifies the start mode of the Windows service.
+It can be one of the following values: Boot, System, Automatic, or Manual.
+For more information, see the [ChangeStartMode method](https://docs.microsoft.com/windows/win32/cimwin32prov/changestartmode-method-in-class-win32-service).
+The default value is `Automatic`.
+
+### delayedAutoStart
+
+This Boolean option enables the delayed start mode if the `Automatic` start mode is defined.
+For more information, see [Startup Processes and Delayed Automatic Start](https://techcommunity.microsoft.com/t5/ask-the-performance-team/ws2008-startup-processes-and-delayed-automatic-start/ba-p/372692).
+
+Please note that this startup mode will not take affect on old Windows versions older than Windows 7 and Windows Server 2008.
+Windows service installation may fail in such case.
+
+```yaml
+delayedAutoStart: false
+```
+
+### depend
+Specify IDs of other services that this service depends on. 
+When service `X` depends on service `Y`, `X` can only run if `Y` is running.
+
+YAML list can be used to specify multiple dependencies.
+
+```yaml
+depend:
+    - Eventlog
+    - W32Time
+```
+
+### log
+
+Optionally set a different logging directory with `logpath` and startup `mode`: append (default), reset (clear log), ignore, roll (move to `\*.old`).
+
+User can specify all log configurations as a single YAML dictionary
+
+```yaml
+log:
+    mode: roll-by-size
+    logpath: '%BASE/log%'
+    sizeThreshold: 10240
+    keepFiles: 8
+```
+
+See the [Logging and error reporting](loggingAndErrorReporting.md) page for more info.
+
+### Arguments
+
+`arguments` element specifies the arguments to be passed to the executable. User can specify all the commands as a single line.
+
+```yaml
+args: arg1 arg2 arg3
+```
+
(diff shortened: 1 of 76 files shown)


src/Core/WinSWCore/Configuration/YamlConfiguration.cs
```diff
@@ -496,19 +513,44 @@ public void LoadEnvironmentVariables()
...
+        public string BaseName => this.Defaults.BaseName;
 
-        public string BasePath => throw new NotImplementedException();
+        public string BasePath => this.Defaults.BasePath;
 
-        public string? ServiceAccountDomain => throw new NotImplementedException();
+        public string? SecurityDescriptor => this.SecurityDescriptorYaml ?? this.Defaults.SecurityDescriptor;
 
-        public string? ServiceAccountName => throw new NotImplementedException();
+        public static TimeSpan ParseTimeSpan(string v)
```
oleg-nenashev, line 527: Code duplication from XML? I would suggest to move it to a separate utility class then

src/Core/WinSWCore/Configuration/YamlConfiguration.cs
```diff
@@ -100,6 +100,9 @@ public class YamlConfiguration : IWinSWConfiguration
         [YamlMember(Alias = "securityDescriptor")]
         public string? SecurityDescriptorYaml { get; set; }
 
+        [YamlMember(Alias = "extensions")]
+        public List<string>? YamlExtensionIds { get; set; }
```
oleg-nenashev, line 104: This one will not work as defined, right? We still need to support extensions somehow

src/Core/WinSWCore/Configuration/YamlConfiguration.cs
buddhikac96: Make sense. will update.

src/Core/WinSWCore/Configuration/YamlConfiguration.cs
buddhikac96: Yes. It is my next plan on YAML Support

examples/sample-allOption.yml
```diff
@@ -0,0 +1,72 @@
+id: myapp
+name: MyApp Service (powered by WinSW)
+description: This service is a service created from a sample configuration
+executable: java
+caption: WinSW
```
NextTurn, line 5: There is no corresponding entry in XML.

examples/sample-allOption.yml
```diff
@@ -0,0 +1,72 @@
...
+#serviceaccount:
+#    domain: YOURDOMAIN
+#    user: useraccount
+#    password: Pa55w0rd
+#    allowservicelogon: yes
+#failureActions:
```
NextTurn, line 11: I don't think `failureActions` is better than `onFailure`.

examples/sample-allOption.yml
```diff
@@ -0,0 +1,72 @@
...
+#securityDescriptor: security descriptor string
+#arguments: >
+#    -classpath 
+#    c:\cygwin\home\kohsuke\ws\hello-world\out\production\hello-world 
+#    test.Main
+#startArguments: start arguments
+#workingdirectory: C:\myApp\work
+priority: Normal
+stopTimeout: 15 sec
+stopParentProcessFirst: false
```
NextTurn, line 30: Use `true` in samples.

examples/sample-allOption.yml
buddhikac96: will change into `onFailure`

examples/sample-allOption.yml
buddhikac96: Its a mistake. Should be removed.

examples/sample-allOption.yml
```diff
@@ -0,0 +1,71 @@
...
+#    - Eventlog
+#    - W32Time
+waitHint: 15 sec
+sleepTime: 1 sec
+#interactive: true
+log:
+#    logpath: '%BASE%\logs'
+    mode: append
+#env:
+#    MY_TOOL_HOME: 'C:\etc\tools\myTool'
```
NextTurn, line 44: - Shall we support a similar syntax for XML in v3?
- Shall we support `name:` and `value:` pairs?

buddhikac96: Sure. I am okay with that! I will update this.

buddhikac96: > * Shall we support a similar syntax for XML in v3

Sorry, I did not get this???

oleg-nenashev: I believe @NextTurn asks whether we should change XML syntax in v3 so that YAML and XML are aligned

buddhikac96: What would be good?? I am open for both. However I already change YAML config into XML syntax. I will push that. But still I can roll back to previous syntax if you wish.

doc/YamlConfigFile.md
```diff
@@ -0,0 +1,299 @@
...
+YAML Configuration file shuold be in following format
+
+Example:
+
+```yaml
+id: jenkins
+name: Jenkins
+description: This service runs Jenkins continuous integration system.
+env:
+    -
```
NextTurn, line 19: Blank lines aren't required.

src/Core/WinSWCore/Configuration/YamlConfiguration.cs
```diff
@@ -315,15 +325,29 @@ public class YamlDownload
...
+            public SC_ACTION_TYPE Type
+            {
+                get
+                {
+                    SC_ACTION_TYPE actionType = this.FailureAction switch
+                    {
+                        "restart" => SC_ACTION_TYPE.SC_ACTION_RESTART,
+                        "none" => SC_ACTION_TYPE.SC_ACTION_NONE,
+                        "reboot" => SC_ACTION_TYPE.SC_ACTION_REBOOT,
+                        _ => throw new Exception("Invalid failure action: " + this.FailureAction)
```
NextTurn, line 343: Don't throw `Exception`.

doc/YamlConfigFile.md
```diff
@@ -0,0 +1,299 @@
...
+```
+
+See the [Logging and error reporting](loggingAndErrorReporting.md) page for more info.
+
+### Arguments
+
+`arguments` element specifies the arguments to be passed to the executable. User can specify all the commands as a single line.
+
+```yaml
+args: arg1 arg2 arg3
```
NextTurn, line 107: As per the discussion in Gitter we don't want abbreviations. Can you confirm @oleg-nenashev?

src/Core/WinSWCore/Configuration/YamlConfiguration.cs
buddhikac96: Which value should be bound to actionType for a invalid input?

NextTurn: Throw `InvalidDataException` ~~or `InvalidEnumArgumentException`~~ instead.

doc/YamlConfigFile.md
buddhikac96: Updated

doc/YamlConfigFile.md
```diff
@@ -0,0 +1,298 @@
...
+    -jar "%BASE%\jenkins.war"
+    --httpPort=8080
+log:
+    mode: roll
+```
+
+## Environment variable expansion
+
+Configuration YAML files can include environment variable expansions of the form `%Name%`. 
+Such occurrences, if found, will be automatically replaced by the actual values of the variables. [Read more. . . ](xmlConfigFile.md#environment-variable-expansion)
```
oleg-nenashev, line 34: Is it really supported at the moment? I do not see it in code though I am looking from phone and might be missing it

doc/YamlConfigFile.md
```diff
@@ -0,0 +1,298 @@
...
+
+## Configuration entries
+
+### id
```
oleg-nenashev, line 38: There is a lot of duplication between this doc and XML. Maybe we should move detailed docs into a separate doc and link them.

It does not block this PR, can be done later. Docs need update for v3 anyway

buddhikac96:
> Looks good overall. Just a few comments. I'd guess.we want to proceed without the extensions implementation here, right?

Yeah. At the moment I am working on that on another branch.

doc/YamlConfigFile.md
buddhikac96: Indeed. I am also thinking its more user friendly. Anyway if we move into `winsw.github.io` then we can present that in much more user friendly way.

doc/YamlConfigFile.md
buddhikac96: I guess we are talking about [this](https://github.com/buddhikac96/winsw/blob/9dd7fb4d93055a582628ce8d963b1016d9715c38/src/Core/WinSWCore/Configuration/YamlConfiguration.cs#L500)??

oleg-nenashev: Nope. Syntax like `  <executable>%BASE%\myExecutable.exe</executable>` in XML when the variable is actually used in other fields. This is what `Environment variable expansion` means

doc/YamlConfigFile.md
```diff
@@ -0,0 +1,298 @@
+# YAML configuration file
+
+This page describes the configuration file, which controls the behavior of the Windows service.
+
+You can find configuration file samples in the [examples](../examples) directory of the source code repository.
+Actual samples are being also published as a part of releases on GitHub and NuGet.
```
NextTurn, line 6:
```suggestion
Actual samples are also being published as a part of releases on GitHub and NuGet.
```

doc/YamlConfigFile.md
NextTurn, line 34: `Read more. . . ` looks weird.

doc/YamlConfigFile.md
```diff
@@ -0,0 +1,298 @@
+# YAML configuration file
+
+This page describes the configuration file, which controls the behavior of the Windows service.
+
+You can find configuration file samples in the [examples](../examples) directory of the source code repository.
+Actual samples are also being published as a part of releases on GitHub and NuGet.
```
NextTurn, line 6:
```suggestion
Actual samples are also being published as part of releases on GitHub and NuGet.
```

/cc @oleg-nenashev

doc/YamlConfigFile.md
buddhikac96: Updated

doc/YamlConfigFile.md
oleg-nenashev: I am also not a native speaker :( 
I think we will be doing a major cleanup for v3 anyway

Issue #297, opened 2019-01-16
Keep Sync between the config file and the service


ricardopolo:
Hello.

We have an auto updatable service, thanks to download feature of winsw. We are able to update everything but the configuration of the service. We can update the 'xml" config file, however the Windows Service Console never pickup changes.

In the config file we have the service description (with the version of the installed service), or configuration about the service that would be great to keep updated. Any ideas to keep it updated whitout manually uninstall and reinstall the service? LIke a way to always overwrrite the service config from the XML?

THanks!!

oleg-nenashev:
There are some fields in WinSW configuration which are actually stored in Windows Registry on the `install` command execution. These fields will not be updated unless the service is reinstalled, and the documentation marks these fields. The rest of configuration fields should be updated on each windows service wrapper restart.

Could you please provide examples of the configuration options which do  not get updated?

ricardopolo:
- Startup Type (automatic, delayed, etc).
- Service Description (this is what I found helpful for show the version in the services console, and automatically update it)

oleg-nenashev:
Yes, bot these fields require the service reinstallation now.
But it may be possible to support dynamic registry update.

NextTurn:
Hopefully this issue is not too stale.

Actually all configuration can be updated without a re-installation, except that some changes won't be reflected on the fly (login info).

My ideas:

Let's ignore the credentials for now due to availability problems.

### How to know when/what to update (see also #439)

1. Store the last write time of the config file. Update all config after the file is updated.
2. Store a copy of the installation config. Compare and update.

### When to update

#### Manual ways

- Add a new command which updates all configuration.
  - No different from `uninstall` + `install` if only the last write time is stored.

#### Automatic ways

1. When executing `start` or `restart` commands.
2. When the service is running.

Which do you prefer?

ricardopolo:
I thin when service is running is better

Issue #256, opened 2017-11-16
pre-start argument


lixiangyu4:
I want to call an external script before starting the service, I only see stopargument in the xml configuration, but there is no startargument configuration
I want to achieve the function is to start the service call script detection configuration properties file port has been modified, if you modify the component configuration file to be changed

PR #607, opened 2020-07-26
Add `refresh` command

NextTurn:
Contributes to #256 
Contributes to #297


oleg-nenashev:
Great addition!

NextTurn:
Note that this doesn't update properties that may require a restart. I will add support for them in the next version.

ricardopolo:
@nxtn @oleg-nenashev any advice or workaround to have a similar option like Refresh in v2?
We need to use it now but 3 is still in beta. Maybe run `install` command again refresh the options?

Thanks

PR #614, opened 2020-07-30
Preshutdown support

### LINKED ISSUE #489: Giving the service more time to stop when Windows restarts
### Feature Request

**Problem statement.**
For services that have a slow shutdown, Windows gives up on the graceful shutdown of the services and terminates them.

**Describe the solution you'd like.**

We would like to be able to utilize preShutdown handler.

See this link: https://www.coretechnologies.com/blog/windows-services/increase-shutdown-time/

> How to extend the preshutdown timeout beyond 3 minutes
> 
> With a simple code change, a preshutdown service can instruct Windows to wait more than the default 3 minutes for the service to end in the preshutdown phase.
> 
> You must add code to call the ChangeServiceConfig2 function with the SERVICE_CONFIG_PRESHUTDOWN_INFO level, and the timeout value in the SERVICE_PRESHUTDOWN_INFO structure set to an appropriate value.
> 
> Here is what the addition looks like in C++ (error checking omitted for brevity):
> 
>    SC_HANDLE hService =	OpenService(...);
>    ...
>    SERVICE_PRESHUTDOWN_INFO info;
>    info.dwPreshutdownTimeout = 5 * 60 * 1000; // 5 minutes
>    ChangeServiceConfig2(hService, SERVICE_CONFIG_PRESHUTDOWN_INFO, &info);
> 
> The new code is best called once when you install/setup your service, but it can be invoked during normal operation as well. Whichever works better in your situation.

Can this somehow be used to make Windows wait longer to restart the service?

**Describe alternatives you've considered.**
We can also increase a registry timeout parameter to give the service more time to restart, but this does not seem to be a very popular suggestion with the system admins who refuse to touch registry settings for our app.

NextTurn:
Fixes #489 

The Event Log service is available during the preshutdown period.

### DIFF
--- a/src/WinSW.Core/Configuration/DefaultSettings.cs
+++ b/src/WinSW.Core/Configuration/DefaultSettings.cs
@@ -55,6 +55,8 @@ namespace WinSW.Configuration
 
         public bool DelayedAutoStart => false;
 
+        public bool Preshutdown => false;
+
         public string[] ServiceDependencies => new string[0];
 
         public bool Interactive => false;
--- a/src/WinSW.Core/Native/Service.cs
+++ b/src/WinSW.Core/Native/Service.cs
@@ -273,6 +273,18 @@ namespace WinSW.Native
         }
 
         /// <exception cref="CommandException" />
+        internal void SetPreshutdownTimeout(TimeSpan timeout)
+        {
+            if (!ChangeServiceConfig2(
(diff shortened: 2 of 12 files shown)


NextTurn:
I sometimes saw a "service did not shutdown properly after receiving a preshutdown control" error. Couldn't figure out why.

PR #615, opened 2020-07-30
Add `customize` command

### LINKED ISSUE #199: Changing Assembly Information for created services
When I create a new service, I would like to change the default Assembly information for the newly created services in winsw.

```
[assembly: AssemblyTitle("Windows Service Wrapper for .NET4")]
[assembly: AssemblyDescription("Allows arbitrary process to run as a Windows service by wrapping it")]
[assembly: AssemblyConfiguration("")]
[assembly: AssemblyCompany("CloudBees, Inc.")]
[assembly: AssemblyProduct("Windows Service Wrapper")]
[assembly: AssemblyCopyright("Copyright 2008-2016 Oleg Nenashev, CloudBees, Inc. and other contributors")]
```

I would hope to add a new XML configuration with the appropriate fields. Would that be possible? If so, I would be happy to send a PR for this feature. If you have any input as to how to make this feature happen, I would like to hear your thoughts.

NextTurn:
Fixes #199

### Usage

```
winsw customize -o <output> --manufacturer <manufacturer>
```

There is currently a limitation on the new name:

```cs
const int minLength = 12;
const int maxLength = 15;
```

/cc @aminjam

### DIFF
--- a/src/WinSW.Core/Native/ResourceApis.cs
+++ b/src/WinSW.Core/Native/ResourceApis.cs
@@ -1,0 +1,38 @@
+﻿#pragma warning disable SA1310 // Field names should not contain underscore
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace WinSW.Native
+{
+    internal static class ResourceApis
+    {
+        internal static readonly IntPtr VS_VERSION_INFO = new IntPtr(1);
+
+        internal static readonly IntPtr RT_VERSION = new IntPtr(16);
+
+        [DllImport(Libraries.Kernel32, SetLastError = true, CharSet = CharSet.Unicode)]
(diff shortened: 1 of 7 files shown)


oleg-nenashev:
Would it also make sense to introduce a new unit test which operates with a WinSW.exe copy? It would not be a full integration test, but it would provide some coverage

NextTurn:
> Ideally the code should explicitly invalidate code signing if present (it would be broken anyway due to the checksum change), and to require users to re-sign the executable on their own if they need signing

I'll make it clear in the command description. Not easy to play with signatures in our codes.

 > Would it also make sense to introduce a new unit test which operates with a WinSW.exe copy?

Certainly. I plan to write tests late later...

PR #629, opened 2020-08-05
Update release Drafter to v5.11.0

oleg-nenashev:


oleg-nenashev:
There are multiple bugfixes which should not affect WinSW. No breaking changes either, just a housekeeping

PR #641, opened 2020-08-08
Support both XML and YAML configuration + Remove support for *.vshost files and folder hierarchy lookup which used to be an undocumented development features

### LINKED ISSUE #630: YAML Configuration Reference and usage guidelines are missing in Readme
YAML Support is a major feature. As a user, I would expect it to be referenced from README. In particular, the following information is missing:

* No reference toy https://github.com/winsw/winsw/blob/master/doc/YamlConfigFile.md
* There is no guidelines about how to enable the feature, neither on the README nor on the feature page. https://github.com/jenkins-infra/jenkins.io/pull/3573 also does not explain how to try out the feature

## Steps to reproduce

* Check out https://github.com/winsw/winsw/commit/a7fb954f76d49177f03b88e7db23964c1323085e
* Go to the README, there are no references to YAML

buddhikac96:
**Description**
Update Windows Service Wrapper to support both XML and YAML configurations. If there are both configurations files, then WinSW will be configured by the XML configuration file.

**Reference**
Fixes #630 

**Updates**
- Update WinSW to support both XML and YAML configurations
- Reference yamlConfiguration.md from ReadMe.md

PS - Add how to use section to Jenkins blog post. (https://github.com/jenkins-infra/jenkins.io/pull/3573)


README.md
```diff
@@ -33,7 +33,7 @@ Binaries are available [here](https://repo.jenkins-ci.org/releases/com/sun/winsw
 
 ## Usage
 
-WinSW is being managed by configuration files: [Main XML configuration file](doc/xmlConfigFile.md) and [EXE configuration file](doc/exeConfigFile.md).
+WinSW is being managed by configuration files: [Main XML configuration file](doc/xmlConfigFile.md), [Main YML configuration file](doc/YamlConfigFile.md) and [EXE configuration file](doc/exeConfigFile.md).
```
NextTurn, line 36: YML -> YAML. `YamlConfigFile.md` should be renamed to `yamlConfigFile.md` to match the naming convention.

README.md
```diff
@@ -59,6 +59,7 @@ User documentation:
 * [Installation guide](doc/installation.md) - Describes the installation process for different systems and .NET versions
 * Configuration:
   * [Main XML configuration file](doc/xmlConfigFile.md)
+  * [Main YAML configuration file](doc/YamlConfigFile.md)
   * [EXE configuration file](doc/exeConfigFile.md)
```
NextTurn, line 63: Could you remove the "EXE configuration file" by the way? We no longer use it.

src/Core/ServiceWrapper/Program.cs
```diff
@@ -658,6 +671,39 @@ private static string ReadPassword()
             }
         }
 
+        private static ConfigType GetConfigType(out string baseName, out DirectoryInfo d)
```
NextTurn, line 674: I don't think we need an enum here. Simply construct the service descriptor and return it.

src/Core/ServiceWrapper/Program.cs
```diff
@@ -658,6 +671,39 @@ private static string ReadPassword()
...
+                }
+
+                if (File.Exists(Path.Combine(d.FullName, baseName + ".yml")))
+                {
+                    return ConfigType.YAML;
+                }
+
+                if (d.Parent is null)
+                {
+                    throw new FileNotFoundException("Unable to locate " + baseName + ".xml " + "or " + baseName + ".yml " + "file within executable directory or any parents");
```
NextTurn, line 700: Maybe a compact format like `$"{baseName}.[xml|yml]"`? Not a blocker.

README.md
buddhikac96: np problem

README.md
```diff
@@ -33,7 +33,7 @@ Binaries are available [here](https://repo.jenkins-ci.org/releases/com/sun/winsw
 
 ## Usage
 
-WinSW is being managed by configuration files: [Main XML configuration file](doc/xmlConfigFile.md) and [EXE configuration file](doc/exeConfigFile.md).
+WinSW is being managed by configuration files: [Main XML configuration file](doc/xmlConfigFile.md) and [Main YML configuration file](doc/yamlConfigFile.md).
```
NextTurn, line 36: YML -> YAML.

src/Core/ServiceWrapper/Program.cs
```diff
@@ -658,6 +659,39 @@ private static string ReadPassword()
             }
         }
 
+        private static IWinSWConfiguration GetConfigType()
```
NextTurn, line 662: The method name needs an update.

src/Core/ServiceWrapper/Program.cs
```diff
@@ -658,6 +659,39 @@ private static string ReadPassword()
...
+        {
+            var executablePath = new DefaultWinSWSettings().ExecutablePath;
+            var baseName = Path.GetFileNameWithoutExtension(executablePath);
+
+            if (baseName.EndsWith(".vshost"))
```
NextTurn, line 667: You could also remove this block by the way. This should not be a part of the main logic.

buddhikac96: Sorry, I don't get it? Don't we need to remove .vshost extension before start searching the file?

NextTurn: No.

oleg-nenashev: It makes sense to keep it for XML though

NextTurn: This was only for testing. The test host today is not *.vshost.exe*.

NextTurn:
We didn't say we support recursive lookup in all versions of the documentation. Again, it was only used to "simplify debugging".

```
// find co-located configuration xml. We search up to the ancestor directories to simplify debugging,
// as well as trimming off ".vshost" suffix (which is used during debugging)
```

buddhikac96:
> The change may still cause incompatibility if there is a `service.xml` file in upper folder, and unrelated `service.yml` in the process root. You could prevent it by putting the files discovery in the same loop.
> 
> At the same time, I do not think such config file discovery behavior should be retailed in v3. IMHO it should be either executable root or explicit config path definition there.

So the final decision is, we don't find the config file recursively and we don't trim the .vshost??

src/Core/ServiceWrapper/Program.cs
```diff
@@ -658,6 +659,26 @@ private static string ReadPassword()
...
+            {
+                return new ServiceDescriptor(baseName, d);
+            }
+
+            if (File.Exists(Path.Combine(d.FullName, baseName + ".yml")))
+            {
+                return new ServiceDescriptorYaml(baseName, d).Configurations;
+            }
+
+            throw new FileNotFoundException($"Unable to locate { baseName }.[xml|yml] file within executable directory or any parents");
```
NextTurn, line 679: "or any parents" still in the string.

oleg-nenashev: nice catch

NextTurn:
@buddhikac96 Don't use pull requests to update your repository, use `git pull` + `git push` instead.

Issue #606, opened 2020-07-26
Seriously think about why users are moving to NSSM


NextTurn:
https://stackoverflow.com/questions/40846356/run-nginx-as-windows-service

It sucks.

losvedir:
I transitioned my app from `nssm` to `WinSW` this week (to be consistent with the rest of my organization), and have now had two production downtimes and am thinking of going back to `nssm`. Maybe I can provide a little feedback.

The biggest *benefit* to WinSW as I see it, is a singular XML file for configuration. This is much easier for us to back up. With `nssm` I kind of paged through their editor interface and wrote down all the settings and backed it up that way.

The biggest drawback is that either 1) it didn't work right and didn't restart my production app, or 2) lack of documentation meaning I did it wrong.

My big issue is that my app crashed (due to a network glitch, not `WinSW`'s fault at all), but then wasn't started up again. It happened once, earlier in the week, and that's when I re-read the config file documentation and noticed the `<onfailure>` key, which I wasn't using. I didn't expect to have to explicitly add something to make the app try to start up again when it crashes. Perhaps I'd been trained by `nssm` defaults, but I thought a Windows Service should just automatically try to periodically restart after it crashes. Isn't that the point of running something as a service?

Anyway, I chalked that up to user error, added the key, *restarted the service via the Windows Services panel*, and went on my way. But then it crashed again late late last night, and stayed down again. So now I'm not confident in `WinSW` and unless I figure out what's going on here, I plan to go back to `nssm`. In particular, I'm wondering if I needed to have done something *via the command line `my_app.exe`* tool for it to pick up the new config settings? From some testing, simply restarting the app via the Windows Services panel was enough to pick up new environment variables, so I assumed it was picking up changes. But maybe not to `<onfailure>`? This gets back to a lack of documentation, perhaps.

oleg-nenashev:
Thanks for the feedback @losvedir !

> I didn't expect to have to explicitly add something to make the app try to start up again when it crashes. 

We could change the defaults in v.3.0 

>  From some testing, simply restarting the app via the Windows Services panel was enough to pick up new environment variables, so I assumed it was picking up changes. But maybe not to <onfailure>?

[...]

PR #648, opened 2020-08-15
Auto refresh

### LINKED ISSUE #256: pre-start argument

NextTurn:
Fixes #256 
Contributes to #297 
Contributes to #606 

Automatically refreshes service properties if the config file is newer when the service is started or a command with write access is executed.

Enabled by default. Specify `<autoRefresh>false</autoRefresh>` to override.

### Known issues

The last write time of the registry key sometimes isn't updated (or is updated abnormally like the followings)... so you may end up with refreshing the service indefinitely.

```console
> WinSW.NET461.exe stop
file     last modified at 2020-08-15 3:39:23
registry last modified at 2020-08-15 3:25:44
03:39:27,053 - Service 'ABCD (ABC)' was refreshed successfully.
03:39:27,073 - Stopping service 'ABCD (ABC)'...
03:39:27,077 - Service 'ABCD (ABC)' has already stopped.

> WinSW.NET461.exe stop
file     last modified at 2020-08-15 3:39:30
registry last modified at 2020-08-15 3:32:17
03:39:33,356 - Service 'ABCDE (ABC)' was refreshed successfully.
03:39:33,376 - Stopping service 'ABCDE (ABC)'...
03:39:33,381 - Service 'ABCDE (ABC)' has already stopped.

> WinSW.NET461.exe stop
file     last modified at 2020-08-15 3:39:30
registry last modified at 2020-08-15 3:32:23
03:40:35,601 - Service 'ABCDE (ABC)' was refreshed successfully.
03:40:35,621 - Stopping service 'ABCDE (ABC)'...
03:40:35,625 - Service 'ABCDE (ABC)' has already stopped.
```

/cc @ricardopolo @losvedir

### DIFF
--- a/src/WinSW/Program.cs
+++ b/src/WinSW/Program.cs
@@ -20,6 +20,7 @@ using log4net.Appender;
 using log4net.Config;
 using log4net.Core;
 using log4net.Layout;
+using Microsoft.Win32;
 using WinSW.Logging;
 using WinSW.Native;
 using WinSW.Util;
@@ -88,7 +89,7 @@ namespace WinSW
                     XmlServiceConfig config = null!;
                     try
                     {
-                        config = CreateConfig(pathToConfig);
+                        config = LoadConfig(pathToConfig);
                     }
                     catch (FileNotFoundException)
                     {
@@ -98,6 +99,9 @@ namespace WinSW
                     InitLoggers(config, enableConsoleLogging: false);
 
                     Log.Debug("Starting WinSW in service mode.");
+
+                    AutoRefresh(config);
+
                     ServiceBase.Run(new WrapperService(config));
                 }),
             };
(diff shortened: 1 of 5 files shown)


src/WinSW/Program.cs
```diff
@@ -986,10 +961,82 @@ static void Elevate(bool noElevate)
                     Environment.Exit(e.ErrorCode);
                 }
             }
+
+            static void AutoRefresh(XmlServiceConfig config)
+            {
+                if (!config.AutoRefresh)
+                {
+                    return;
```
oleg-nenashev, line 969: Put a message to the log? Could help to raise awareness about the feature

oleg-nenashev (commented):
For diagnosability purposes, it would be great to compare with existing configs and to log messages when changes are detected. It could be done in future iterations as an improvements.

Please update documentation before merging

src/WinSW/Program.cs
NextTurn: Makes sense. Which log level will be appropriate?

oleg-nenashev: INFO?

NextTurn: I enabled this feature by default... Any problems?

ricardopolo:
Is refresh command available in v2? We need to use it right now but 3 looks in Alpha :(

Issue #665, opened 2020-08-28
Runaway Process Killer Yaml: StopParentFirst uses inconsistent name


offa:
The Yaml configuration of Runaway Process Killer starts with an upper case, while all other options – including the xml counterpart – start lower case.

```yml
StopParentFirst: true
```

This is really minor, but breaks consistency.

## Steps to reproduce

Create a Yaml configuration using the option ([docs](https://github.com/winsw/winsw/blob/master/doc/extensions/runawayProcessKiller.md)).

## Environment

* WinSW version: 2.10.1
* WinSW package type: all
* Windows version: all
* Wrapped executable and version: all

## Possible Solution

Change to lower case.

NextTurn:
`stopTimeOut` in YAML is also inconsistent with `stopTimeout` in XML.

buddhikac96:
@offa  Thank you so much for creating this.

oleg-nenashev:
YAML support is in Preview, so we can safely fix it in the next releases. It will be a minor breaking change, though even that can be prevented if YamlDotNet supports multiple aliases

PR #678, opened 2020-09-02
Fix inconsistencies in YAML config 

buddhikac96:
Previously there were differences between XML and YAML configurations. Mostly case differences. This PR fix that.

Fixes   https://github.com/winsw/winsw/issues/665


oleg-nenashev:
CC @NextTurn . Do we really want to fully align XML and YAML? Camel Case notation looks better IMHO.

NextTurn (changes requested):
Camel case is definitely better.

Issue #605, opened 2020-07-26
Add configuration samples for frequently wrapped apps


NextTurn:
Including but not specific to:

- Nginx
- ...

PR #692, opened 2020-10-06
Sample with node http-server

salem84:
Hi, I have created a sample configuration XML to start [Node Http-Server](https://www.npmjs.com/package/http-server), based on #605 
I have added some comments to change default port and public folder

### DIFF
--- a/samples/node-http-server.xml
+++ b/samples/node-http-server.xml
@@ -1,0 +1,51 @@
+<!--
+  MIT License
+
+  Copyright (c) 2008-2020 Kohsuke Kawaguchi, Sun Microsystems, Inc., CloudBees,
+  Inc., Oleg Nenashev and other contributors
+
+  Permission is hereby granted, free of charge, to any person obtaining a copy
+  of this software and associated documentation files (the "Software"), to deal
+  in the Software without restriction, including without limitation the rights
+  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+  copies of the Software, and to permit persons to whom the Software is
+  furnished to do so, subject to the following conditions:
+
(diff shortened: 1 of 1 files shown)


samples/node-http-server.xml
```diff
@@ -0,0 +1,51 @@
+<!--
+  MIT License
+
+  Copyright (c) 2008-2020 Kohsuke Kawaguchi, Sun Microsystems, Inc., CloudBees,
```
oleg-nenashev, line 4: If you want to keep the license header, please update it so that it refers you or just WinSW Contributors + keep the right date. Also, IMHO the license header can be just deleted so that visitors see the actual content

oleg-nenashev (changes requested):
Thanks for the pull request! Please address the license comment. It would be also great to somehow reference the file from documentation, e.g. by adding a "samples" section in the config xml documentation

samples/node-http-server.xml
salem84: I have copied from other samples, but I agree with you! it's better delete it

salem84:
>  It would be also great to somehow reference the file from documentation, e.g. by adding a "samples" section in the config xml documentation

Adding a bullet list in `samples` paragraph in this page https://github.com/winsw/winsw/blob/v3/docs/developer/project-structure.md ?
Or do you think it's better in another section?

NextTurn:
> Adding a ~~bullet list~~ table in `samples` paragraph in this page https://github.com/winsw/winsw/blob/v3/docs/developer/project-structure.md ?

Would be better to add a *README.md* with this table in the *samples* directory. The "filename" column could be replaced with display names.

PR #694, opened 2020-10-07
Leave STDOUT/STDERR open if not redirected

NextTurn:


oleg-nenashev (approved):
LGTM, also thanks for generalizing the code

PR #719, opened 2020-11-15
Redirect I/O of elevated process

nxtn:


NextTurn:
This patch doesn't compile in earlier .NET versions.

PR #728, opened 2020-12-07
Ensure basic event logging

NextTurn:
Similar to #725. If the config file can't be loaded at all, writes a log entry using source "Windows Service Wrapper".


oleg-nenashev (approved):
Looks good!

PR #730, opened 2020-12-07
Rename framework-dependent deliverables

NextTurn:


oleg-nenashev (approved):
It cleans up the code a bit, so why not? Personally I still have preference to have explicit names so that `TargetFramework` values change does not come as a surprise for the users.

NextTurn:
For a self-contained executable the runtime version is transparent to users, so they would not need to care about whether it's based on .NET Core 3, .NET 5 or even an AOT-compiled runtime.

PR #780, opened 2021-01-21
Update serviceaccount field in sample complete.xml

Zing22:
The field `<serviceaccount>` seems like having a different structure, which not contains `domain` and `user`, but only a `username`.

### DIFF
--- a/samples/complete.xml
+++ b/samples/complete.xml
@@ -65,8 +65,7 @@ Their modification will not take affect without the service re-installation.
   -->
   <!--
   <serviceaccount>
-    <domain>YOURDOMAIN</domain>
-    <user>useraccount</user>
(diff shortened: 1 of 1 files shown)


cbdchang:
I used the following,

```
  <serviceaccount>
    <user>useraccount</user>
    <password>Pa55w0rd</password>
    <allowservicelogon>true</allowservicelogon>
  </serviceaccount>
```

to successfully install the service as a local user account. The instruction in `docs/xml-config-file.md` didn't work for me (the installation used the LocalSystem account) but the example in `complete.xml` did work for me. I think the doc should be updated to provide both uses.

nxtn:
closing and reopening for a new CI build

nxtn (approved):
Thanks!

PR #784, opened 2021-01-25
Fixed bug constructing command-line for elevated process in .NET 5

yeurch:
When elevating with .NET 5.0, the constructed command-line is incorrect, due to spacing.  For example, when elevating the `start` command, we get:

```
WinSW.exe /elevated  - - -start
```

when what we actually want is

```
WinSW.exe /elevated - - - start
```


oleg-nenashev (approved):
thanks for catching it

PR #793, opened 2021-02-17
Ensure we close correct handle after checking if service exists

yeurch:
Previously, this code would close the Service Control Manager (SCM)
handle instead of the service handle, meaning:

- any future calls to the SCM would fail due to invalid handle
- a handle to the service was left open in error

Also, reduce the access rights given to the handle, as we don't need
ALL rights just to check if the service exists.


yeurch:
Note that this bug also exists in `master`, so the fix should probably be ported there too.

nxtn (approved):
Thanks for spotting!

PR #866, opened 2021-09-10
Update issue templates

benwa:
Using https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms

image


.github/ISSUE_TEMPLATE/4-question.yml
```diff
@@ -0,0 +1,10 @@
+name: ❓ Need help
```
benwa, line 1: This is probably better handled with Gitter and the Discussions feature than an Issue

oleg-nenashev: Yes, ideally to be moved.

oleg-nenashev (approved):
IIUC the beta feature needs to be enabled first before the new UI will be operational

PR #879, opened 2021-10-30
Add config element to load environment variables from a file

sbaeumlisberger:
Hi, I implemented a config option to load environment variables from a file.


odin568:
Would love to have that merged for v2 - exactly what I was looking for. 
Reason is I want to generate the configuration as I have around 30 env variables to be defined. Having those in an env file would have the benefit to have a static config file and just generate the variable parts agnostic to any general winsw config layout changes.

PR #930, opened 2022-05-01
Fix incorrect quotation marks

freezingDaniel:
Seems like someone's editor replaced some quotation marks.
This caused issues when I tried to use the sample as a template. :)


freezingDaniel:
force pushed to trigger recheck after rebase.

oleg-nenashev (approved):
Thanks!

PR #943, opened 2022-05-24
Update xml-config-file.md - Add missing `/` to service account block.

queglay:
Add missing / to service account block.


nxtn (approved):
Thanks!

PR #964, opened 2022-09-01
fix: wrong yaml schema for `logPath`

viceice:
https://github.com/winsw/winsw/blob/9bfc034807a23e8de522efbad0a416b683557d4c/src/WinSW.Core/Configuration/YamlServiceConfig.cs#L376

Needs to be camlcase.


viceice:
Any chance to get it merged, so others don't make same mistake?

nxtn (approved):
Thanks!

PR #998, opened 2023-01-19
Fix build pipelines

nxtn:


nxtn:
microsoft/azure-pipelines-tasks#16088

PR #1010, opened 2023-02-01
fix unhandled exception when signaling service stoped

viceice:
should fix the error occured in
- #980

So the process can properly exit
https://github.com/winsw/winsw/blob/6cf303c1d3fbe1069d95af230b8efa117d29cdf2/src/WinSW/WrapperService.cs#L419-L424

### DIFF
--- a/src/WinSW/WrapperService.cs
+++ b/src/WinSW/WrapperService.cs
@@ -391,10 +391,14 @@ namespace WinSW
 
         private void SignalStopped()
         {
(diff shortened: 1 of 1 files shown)


viceice:
@nxtn Can you please review / merge and release this fix?

oleg-nenashev (commented):
I would do proper error propagation there, but I need to look into the code first.
Sorry for a multi-year folllow up delay

PR #1028, opened 2023-03-14
Update process tree termination to be bottom up

### LINKED ISSUE #1027: Error Stopping Service
I have wrapped a custom application with winsw. My application creates child processes which then go on to create child processes of their own. When the windows service is stopped, some of the child processes get orphaned and thus cannot be stopped.

## Steps to reproduce

1. Create an application that creates at least two levels of child processes below the main application.
2. Wrap said application with winsw
3. Install the service
4. Start the service
5. Wait for your two levels of child processes to spawn
6. Stop the service

## Environment

* WinSW version: v3.0.0-alpha.11
* WinSW package type: x64
* Windows version: Server 2019
* Wrapped executable and version: internal project

## Possible Solution

It appears winsw takes a top down approach to stopping the process tree. I believe a bottom up approach will solve the issue.

I have attached the appropriate logs for my issue.
[myapp.wrapper.log](https://github.com/winsw/winsw/files/10971901/myapp.wrapper.log)

valainisgt:
Fixes #1027

### DIFF
--- a/src/WinSW.Core/Util/ProcessExtensions.cs
+++ b/src/WinSW.Core/Util/ProcessExtensions.cs
@@ -17,8 +17,6 @@ namespace WinSW.Util
 
         public static void StopTree(this Process process, int millisecondsTimeout)
         {
-            StopPrivate(process, millisecondsTimeout);
-
             foreach (var child in GetChildren(process))
             {
                 using (child.Process)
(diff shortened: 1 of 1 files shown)


bmoore117:
bumping this issue, I believe this to be the correct fix for the exact same issue I am experiencing as well.

oleg-nenashev (changes requested):
I think we will need to make it configurable. The problem with bottom-up termination is that it is likely to provoke all fail-over process recreation by higher-level processes, resulting in #1027 again. A reliable process tree termination, by default, should start from the top and give an opportunity to the process to properly terminate its children. bottom-up cleanup may happen only after it.

I will try to submit an alternative patch that makes the process more flexible

PR #1096, opened 2024-09-14
Fix typos and update documentation for camelCase yaml configuration

nhermosilla14:
There were some issues regarding the documentation for the yaml config file. I checked the json schema and eventually noticed many in the markdown doc, and a couple in the example. This hopefully solves those issues.


oleg-nenashev (approved):
This change is correct as per https://github.com/winsw/winsw/blob/v2/src/WinSW.Core/Configuration/YamlServiceConfig.cs#L31 . Camel Case naming is used in the deserializer by default, and it is not configurable. IMHO it won't be configurable at this point, even if it differs from XML

PR #1098, opened 2024-10-02
covering missing service commands + upgrade unit tests packages

alexkaos87:


oleg-nenashev:
Thank you! I will see whether it could be re-integrated after the recent changes

PR #1135, opened 2026-01-05
docs: add troubleshooting guide

### LINKED ISSUE #1113: Add a Troubleshooting Section to Docs
I will create a PR with this soon, but using this so i can mention it in other tickets and have something i can use for reference.

### The system cannot find the file specified.

This can be caused by multiple reasons:
* Using the executable name and the executable is not in the path.
  * Solutions: 
    * Use full path
    * Add executable directory to the PATH of the User that runs the service or to the Machine PATH.
* The executable is in the PATH, but it is a bat file.
  * Solution: Use the full path with extension or use the filename with extension
  * Why? It looks like `Process.Start` with `UseShellExecute=False` cannot resolve bat files in the path without the extension.
  * [See working example here](https://github.com/bmarroquin/winsw_notes/tree/main/bat_as_service)

### The executable is found in powershell/cmd but not WinSW.
Same as `The system cannot find the file specified.`  The executable is in the path of the current user, but it might be a bat file or not in the System path. 

### I can't access/do X while running through WinSW
Make sure the user that runs the service has permissions to X.

There are limitations as to what a service can do on Windows, like displaying windows or pop ups. Try making the service interactive and see if it works. Interactive services are not recommended by Microsoft anymore and might not work correctly. If you need to do something that is limited, consider running the software in the background and using the task scheduler to start it automatically.  

### Not working on Windows 7
Install .net framework 4.6.1

meiranzheng:
Fixes #1113.

Adds a new `docs/troubleshooting.md` page with common startup/permission issues and links it from `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ Most commands require Administrator privileges to execute. WinSW will prompt for
 * Use cases:
   * [Self-restarting services](docs/self-restarting-service.md)
   * [Deferred file operations](docs/deferred-file-operations.md)
+  * [Troubleshooting](docs/troubleshooting.md)
 
 ## Contributing
 
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -3,6 +3,7 @@
 If your service fails to start, first check:
 
 - The stdout/stderr logs configured via the [`log` element](logging-and-error-reporting.md#logging)
+- The wrapper log (`<config-file-name>.wrapper.log`) in the log directory (respects `<logpath>`; defaults to the configuration file directory)
 - The Windows Event Log (WinSW reports more details there when running as a service)
 
 ## The system cannot find the file specified.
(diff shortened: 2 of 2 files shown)


meiranzheng:
Small follow-up: the troubleshooting page now also mentions the wrapper log (`<config-file-name>.wrapper.log`) alongside stdout/stderr + Event Log, which helps diagnose start failures / error 1067.

meiranzheng:
Friendly ping: this is doc-only and currently `mergeable_state=clean`.

It adds a troubleshooting page (Fixes #1113) and mentions the wrapper log (`<BaseName>.wrapper.log`) alongside stdout/stderr + Event Log for diagnosing start failures.

When you have a moment, could someone take a quick look? Thanks!

meiranzheng:
Friendly ping: this is still clean/mergeable. Happy to tweak wording or add an extra example if you'd like — thanks!

oleg-nenashev (approved):
Thank you. This is definitely a nice start for this documentation page

Issue #1136, opened 2026-01-17
WinSW cannot handle clean program exits when running as a restricted service user


BobVul:
## Steps to reproduce

1. Create a service that exits with exitcode 0:

  ```xml
    <executable>C:\Windows\System32\cmd.exe</executable>
    <arguments>/c "echo exiting cleanly &amp; exit 0"</arguments>
  ```

2. Install the service (`winsw.exe install`)
3. Create a normal unprivileged user
4. Go to service properties and change service to run as the new user. User will be granted log on as service rights automatically.
5. Start the service

### Actual results

An exception is thrown and the service remains 'running'.

```
2026-01-17 18:59:53,274 ERROR - Thread failed unexpectedly
WinSW.CommandException: Failed to open the service control manager database. Access is denied.
 ---> System.ComponentModel.Win32Exception (5): Access is denied.
   --- End of inner exception stack trace ---
   at WinSW.Native.Throw.Command.Win32Exception(String message)
   at WinSW.Native.ServiceManager.Open(ServiceManagerAccess access)
   at WinSW.WrapperService.SignalStopped()
   at WinSW.WrapperService.<>c__DisplayClass31_0.<StartProcess>g__OnProcessCompleted|0(Process proc)
   at WinSW.Util.ProcessHelper.<>c__DisplayClass5_0.<StartProcessAndCallbackForExit>b__0()
   at WinSW.Util.ProcessHelper.<>c__DisplayClass6_0.<StartThread>b__0()
```

### Expected results

No errors and the service is marked as stopped

## Environment

* WinSW version: 2.12.0
* WinSW package type: x64
* Windows version: Windows 11 24H2, Windows Server 2019, etc.
* Wrapped executable and version: Any, sample with cmd.exe

## Possible Solution

I assume this is because WinSW tries to open the service manager when the process stops:

https://github.com/winsw/winsw/blob/4640393a125c3dc1deb0eb430e6b7f662a7cae29/src/WinSW/WrapperService.cs#L392-L398

This is not allowed when the service is running as a restricted user, which should not have permissions to manage services in general.

This appears to be caused by the backport attempt in #751, and I believe it also affects the same code in v3 from #426.

[...]

PR #1141, opened 2026-01-29
WrapperService: open SCM with Connect in SignalStopped

meiranzheng:
Refs #1136

### Context
Issue #1136 reports an `Access is denied` error when WinSW tries to report a clean child-process exit (`ExitCode == 0`) via `WrapperService.SignalStopped()` while running under a restricted service account.

`SignalStopped()` currently calls `ServiceManager.Open()` (defaults to `ServiceManagerAccess.All`), which can require elevated SCM permissions.

### Change
- Open the SCM with `ServiceManagerAccess.Connect` in `SignalStopped()`.

### Validation
- `dotnet build src/WinSW.Core/WinSW.Core.csproj -c Release`

### Notes
- This is a minimal change on the `v3` branch; if you want a `v2` backport as well, I can prepare it.

### DIFF
--- a/src/WinSW.Core/WrapperService.cs
+++ b/src/WinSW.Core/WrapperService.cs
@@ -489,7 +489,7 @@ namespace WinSW
 
         private void SignalStopped()
         {
-            using var scm = ServiceManager.Open();
+            using var scm = ServiceManager.Open(ServiceApis.ServiceManagerAccess.Connect);
             using var sc = scm.OpenService(this.ServiceName, ServiceApis.ServiceAccess.QueryStatus);
 
             sc.SetStatus(this.ServiceHandle, ServiceControllerStatus.Stopped);


Neustradamus:
@fuguiKz: Nice contribution :)

meiranzheng:
Update from the reporter: they tested the equivalent v2 branch build (net461) with the repro and confirmed it works now.

Ref: https://github.com/winsw/winsw/issues/1136#issuecomment-3846946047

They couldn’t validate v3 yet due to an unrelated startup crash in their environment, but this v3 change is the same minimal SCM-permission adjustment. CI is green; ready for review/merge.

oleg-nenashev (approved):
I am not too concerned about v3, given its pre-release state. On my local build the WinSW v3 seems to work for 0-exit-code process after this change

PR #1149, opened 2026-02-09
Docs: fix SharedDirectoryMapper YAML sample

meiranzheng:
### Motivation
The YAML example in `doc/extensions/sharedDirectoryMapper.md` currently uses:
- `label: N` / `label: M` (missing the trailing `:`)
- `uncpath` (lowercase)

But the implementation expects:
- drive labels like `N:` / `M:`
- the YAML key `uncPath` (capital `P`)

This mismatch can lead to confusing runtime failures for users who copy/paste the sample.

### Change
- Fix the YAML sample to use `label: N:` / `label: M:` and `uncPath`.

### Validation
- Doc-only change.

### DIFF
--- a/doc/extensions/sharedDirectoryMapper.md
+++ b/doc/extensions/sharedDirectoryMapper.md
@@ -51,11 +51,11 @@ extensions:
       settings:
           mapping:
               - enabled: false
-                label: N
-                uncpath: \\UNC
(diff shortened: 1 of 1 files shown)


oleg-nenashev (approved):
yup, same as elsewhere

PR #1152, opened 2026-02-15
SharedDirectoryMapper: make YAML parsing more tolerant

meiranzheng:
## Motivation
When configuring `SharedDirectoryMapper` via YAML, it’s easy to run into runtime/config errors due to:

- key casing differences (`uncpath` vs `uncPath`) because YAML settings are deserialized into `Dictionary<object, object>` and accessed case-sensitively
- drive labels missing the trailing `:` (e.g. `label: N`), which then fail when passed to `WNetAddConnection2`
- non-string scalar values (e.g. YAML booleans) causing invalid casts

## Changes
- Make `SharedDirectoryMapperConfig.FromYaml(...)` resolve required keys (`enabled`, `label`, `uncPath`) case-insensitively.
- Accept non-string scalar values by converting via `ToString()`.
- Normalize single-letter drive labels (`N` -> `N:`).
- Add unit tests covering the above YAML behaviors.

## Testing
- `dotnet test src/WinSW.Tests/WinSW.Tests.csproj -c Release -f net7.0-windows --filter FullyQualifiedName~winswTests.Extensions.SharedDirectoryMapperConfigTest`

### DIFF
--- a/src/WinSW.Plugins/SharedDirectoryMapperConfig.cs
+++ b/src/WinSW.Plugins/SharedDirectoryMapperConfig.cs
@@ -40,16 +40,82 @@ namespace WinSW.Plugins
             if (yamlObject is not Dictionary<object, object> dict)
             {
                 // TODO : throw ExtensionExeption
-                throw new InvalidDataException("SharedDirectoryMapperConfig config error");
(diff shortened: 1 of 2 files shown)


oleg-nenashev (approved):
Thank you! Looks good to me,  and it's nice to have some productization for the code

PR #1159, opened 2026-03-05
move target to dotnet10, multitargeting removed

mtren:
From the latest v2 branch, I changed the dotnet version in the csproj files to 
net10.0-windows, and removed the old targets net20;net40;net461;net7.0-windows. Because net461 and net7.0 are not supported.
Because the projects are no longer multitargeting, I also removed some compile directives that are no longer needed.
Some package References are updated to newer versions.

The exe now only has a dependency on NET 10.
The configured build now fails, because I don't know how to change the build system to not try to build for net7-windows and so on.


oleg-nenashev:
Would it make sense to actually keep WinSW v2 as is, and to update only v3? 
Keeping old versions seems justified to me, because it might be a new major release otherwise

mtren:
I didn't want to change the functionality or build on unreleased versions.
2.12 is the current release and it works. I simply updated the platform and dependencies to make them compatible with current systems.

PR #1160, opened 2026-03-05
Fix misleading XML encoding declaration in the ConfigXmlBuilder test class

### LINKED ISSUE #1158: Misleading XML encoding declaration in ConfigXmlBuilder test utility
**Summary**

While reviewing the project I noticed a small issue in the test utility.

In "src/WinSW.Tests/Util/ConfigXmlBuilder.cs" the XML declaration is generated as:

<?xml version="1.0" encoding="utf-8"?>

However, the XML is constructed as an in-memory .NET string. Since .NET
strings are internally UTF-16, the encoding declaration can be misleading
and does not reflect the actual representation of the XML content.

## Steps to reproduce

1. Open "src/WinSW.Tests/Util/ConfigXmlBuilder.cs"
2. Locate the following code:

str.Append("<?xml version=\"1.0\" encoding=\"utf-8\"?>\n");

The generated XML declaration includes encoding="utf-8" even though
the XML is built as an in-memory string.

Environment

WinSW version: current main branch

WinSW package type: source build

Windows version: N/A

Wrapped executable and version: N/A

Possible Solution

Remove the encoding="utf-8" attribute from the XML declaration and keep:

<?xml version="1.0"?>

Optionally, a small unit test could be added to ensure the encoding attribute
is not included in the generated XML in the future.

If this approach is acceptable, I would be happy to prepare a pull request.

1PetarVujakovic:
Fixes #1158

In "src/WinSW.Tests/Util/ConfigXmlBuilder.cs" the XML declaration was generated as:

"<?xml version="1.0" encoding="utf-8"?>"

Since the XML is constructed as an in-memory .NET string, specifying the encoding
can be misleading (the string is internally UTF-16).

This change removes the encoding attribute and keeps:

"<?xml version="1.0"?>"

### DIFF
--- a/.github/workflows/changelog-management.yml
+++ b/.github/workflows/changelog-management.yml
@@ -6,6 +6,6 @@ jobs:
   update_draft_release:
     runs-on: ubuntu-latest
     steps:
-      - uses: release-drafter/release-drafter@v6.2.0
+      - uses: release-drafter/release-drafter@v6.4.0
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/src/WinSW.Tests/Util/ConfigXmlBuilder.cs
+++ b/src/WinSW.Tests/Util/ConfigXmlBuilder.cs
@@ -59,8 +59,9 @@ namespace WinSW.Tests.Util
             var str = new StringBuilder();
             if (this.PrintXmlVersion)
             {
-                // TODO: The encoding is generally wrong
-                str.Append("<?xml version=\"1.0\" encoding=\"utf-8\"?>\n");
+                // XML is constructed as an in-memory .NET string,
+                // so specifying encoding here can be misleading.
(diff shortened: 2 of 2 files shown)


oleg-nenashev (changes requested):
I am not sure about it, because there should be file I/O roundtrips too. Of course, the encoding is not needed for most tests, and it could be added only when file IO actually happens.

@1PetarVujakovic  I suggest making it parameterizable via a property, with no encoding added by default

1PetarVujakovic:
> I am not sure about it, because there should be file I/O roundtrips too. Of course, the encoding is not needed for most tests, and it could be added only when file IO actually happens.
> 
> @1PetarVujakovic I suggest making it parameterizable via a property, with no encoding added by default

Thanks for the suggestion! I've updated the PR so the XML encoding declaration is now configurable via a property and omitted by default.

oleg-nenashev (approved):
Thank you!

PR #1195, opened 2026-03-25
Add keepFiles support for time-based log rolling

Maoo690:
Added <keepFiles> option to "roll-by-time" log mode, allowing configuration of how many rolled log files to retain. Updated documentation, XmlServiceConfig parsing, and TimeBasedRollingLogAppender to support and enforce file retention. Old log files are now purged automatically when the limit is exceeded.


oleg-nenashev (commented):
The change looks good overall. If you could also add a smoke test for this feature, it would be great. Happy to merge it afterwards

Maoo690:
> The change looks good overall. If you could also add a smoke test for this feature, it would be great. Happy to merge it afterwards

Thanks for the reminder! I have added 2 unit tests to verify that the purge file function works correctly. 
Please check again. 🙏

image

oleg-nenashev (approved):
Thank you! Let's integrate it

